PR #13, opened 2026-04-01
fix: WAL catch-up regression + response frame size limit

### LINKED ISSUE #11: WAL segments not queryable — only memtable and sealed TS partitions visible to queries
## Summary

After the query engine fix in PR #10, queries now correctly read both the active memtable and sealed TS partitions. However, data sitting in WAL segments (`.seg` files) is not visible to queries. During sustained ingest, the WAL→TS flush pipeline cannot keep up, leaving the majority of ingested data in WAL segments that are durable but invisible to the query engine.

At 100M rows ingested, only ~20M were queryable — the remaining ~80M were in WAL segments.

## Environment

- **NodeDB version**: v0.0.0-beta.1 (commit `0a60295` on branch `fix/timeseries-query-engine`)
- **OS**: Linux 6.17.0 / 8GB RAM / 4 CPU
- **Workload**: 100M rows high-cardinality DNS telemetry via ILP

## Where data lives after 100M ingest

| Location | Rows | Queryable? |
|---|---|---|
| Active memtable | ~60K | Yes |
| Sealed TS partitions (30) | ~20M | Yes (fixed in PR #10) |
| WAL segments | ~80M | **No** |
| **Total ingested** | **100M** | **~20M visible (20%)** |

## How to observe

1. Ingest 10M rows via ILP at sustained rate (~150K/s)
2. Check data layout:

```
$ ls ~/.local/share/nodedb/ts/dns_hammer/ts-* | wc -l
1         ← only 1 sealed partition (~1M rows)

$ ls ~/.local/share/nodedb/wal/*.seg | wc -l
7         ← 7 WAL segments (~8M rows)
```

3. Query:

```sql
SELECT COUNT(*) FROM dns_hammer
→ {"count_all": 2019541}    ← ~2M (1 partition + memtable), not 10M
```

The remaining ~8M rows are durably stored in WAL `.seg` files but never make it to the query path.

## Progression over 100M ingest

| Checkpoint | Ingested | Sealed Partitions | Queryable Rows | Visibility |
|---|---|---|---|---|
| 10M | 10,000,000 | 1 | ~2,019,541 | 20% |
| 50M | 50,000,000 | 9 | ~9,999,971 | 20% |
| 70M | 70,000,000 | ~15 | ~14,270,653 | 20% |
| 100M | 100,000,000 | 30 | ~20,812,443 | 21% |

The ratio stays around 20% — suggesting the WAL→TS flush pipeline runs at roughly 1/5 the ingest rate.

## Impact

- Users see stale/incomplete query results during and after bulk ingest
- `COUNT(*)` returns a fraction of actual data, which is confusing
- No way to manually trigger WAL→TS flush or check flush progress
- Comparison with competitors is unfair — TimescaleDB and InfluxDB make all ingested data immediately queryable

## Expected behavior

One or more of:

1. **WAL segments should be queryable** — the query engine could scan WAL data in addition to memtable + sealed partitions (read path, not requiring flush)
2. **WAL→TS flush should keep up with ingest** — if the flush pipeline processed WAL segments into TS partitions faster, the gap would be smaller
3. **Manual flush command** — `FLUSH TIMESERIES dns_hammer` to force WAL→TS conversion
4. **Visibility indicator** — `SHOW TIMESERIES STATUS` showing ingested vs queryable row counts so users know data is pending

## Comparison context

In our head-to-head benchmarks:
- **TimescaleDB**: 100% of ingested rows immediately queryable after COPY commits
- **InfluxDB**: 100% queryable after write acknowledgment
- **NodeDB**: ~20% queryable during sustained ingest, remainder stuck in WAL

This makes direct query performance comparison difficult — NodeDB's query times look great, but it's querying 1/5 of the data that competitors query.

### LINKED ISSUE #12: GROUP BY on high-cardinality columns fails with response frame too large
## Summary

GROUP BY queries on columns with very high cardinality (50K+ unique values) fail with a `response frame too large` error. The query engine correctly computes all groups, but the serialized JSON response exceeds the wire protocol's maximum frame size.

## Environment

- **NodeDB version**: v0.0.0-beta.1 (commit `e3c3122` on `main`)
- **OS**: Linux 6.17.0 / 8GB RAM / 4 CPU
- **Collection**: Timeseries with ~5.8M visible rows, 50K+ unique `qname` values

## Reproduction

```sql
SELECT qname, COUNT(*) FROM dns_wal_test GROUP BY qname
```

**Result**:

```
ERROR (statement 1): [NDB-9000] internal error: response frame too large: 21875915
```

The response would be ~21MB of JSON (50K+ rows with qname strings + counts).

## Other GROUP BY queries work fine

```sql
SELECT qtype, COUNT(*) FROM dns_wal_test GROUP BY qtype       -- 8 rows, 448ms OK
SELECT rcode, COUNT(*) FROM dns_wal_test GROUP BY rcode        -- 4 rows, 527ms OK
SELECT client_ip, COUNT(*) FROM dns_wal_test GROUP BY client_ip -- 9,997 rows, 659ms OK
```

The issue only appears when the result set is large enough that the serialized JSON exceeds the frame limit.

## Context

This was previously masked by Issue #9 (GROUP BY capped at 10K rows). Now that the cap is removed, queries can produce arbitrarily large result sets that exceed the wire protocol frame size.

In the previous build (commit `0a60295`), GROUP BY qname successfully returned 212,376 rows — suggesting the frame limit may be inconsistent between builds, or the current build has a stricter limit.

## Suggested fixes

1. **Streaming/chunked response** — send results in multiple frames rather than buffering the entire response into a single frame
2. **Configurable frame limit** — allow users to increase the max frame size for large result sets
3. **Server-side pagination** — if the result exceeds the frame limit, automatically apply a LIMIT and return a `has_more` indicator
4. **Implicit LIMIT warning** — if no LIMIT clause is provided on a GROUP BY with high cardinality, warn the user or apply a configurable default

## Impact

Users cannot run GROUP BY on high-cardinality columns (common in DNS/observability workloads):
- `GROUP BY qname` — domain name breakdowns (50K-500K+ unique values)
- `GROUP BY client_ip, qname` — cross-dimension analysis
- `GROUP BY src_ip` in network flow data — millions of unique IPs

Workaround: Use `LIMIT` to constrain results, but this returns arbitrary groups, not necessarily the top-N (ORDER BY + LIMIT would be needed for top-N, which may also hit the frame limit depending on cardinality).

farhan-syah:
## Summary

- **Fix WAL catch-up memory regression** (#11): paginated mmap replay (512 records/cycle) replaces loading entire WAL into memory, reducing per-cycle allocation from hundreds of MB to bounded O(page_size)
- **Fix WAL catch-up CPU regression** (#11): LSN-based deduplication via new `wal_lsn` field on `TimeseriesOp::Ingest` prevents double-ingest of records already processed by live ingest path. Backpressure gate tightened from 75% to 50% SPSC utilization
- **Fix WAL catch-up stall after ingest stops** (#11): idle memtable flush in `maybe_run_maintenance()` flushes all non-empty timeseries memtables after 5 seconds of ingest inactivity, ensuring 100% data visibility
- **Fix GROUP BY on high-cardinality columns** (#12): timeseries scan and aggregate handlers now chunk large result sets via partial responses. Native protocol server splits oversized frames; client reassembles transparently

## Test plan

- [x] `cargo test` — 3613 passed, 0 failed
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean
- [ ] Reproduce #12: `SELECT qname, COUNT(*) FROM dns_wal_test GROUP BY qname` with 50K+ groups should return all rows without frame size error
- [ ] Reproduce #11: ingest 10M rows, stop ingest, verify COUNT(*) reaches 100% within ~10 seconds
- [ ] Verify RSS stays under 1.1GB during sustained ingest (was 1.4GB before fix)
- [ ] Verify ingest rate recovers to ~150K+/s (was 107K/s before fix)

Closes #11
Closes #12

### DIFF
--- a/nodedb-types/src/config/tuning/mod.rs
+++ b/nodedb-types/src/config/tuning/mod.rs
@@ -66,7 +66,7 @@ mod tests {
         assert_eq!(parsed.kv.expiry_reap_budget, 1024);
         assert_eq!(parsed.bridge.slab_page_size, 64 * 1024);
         assert_eq!(parsed.network.default_deadline_secs, 30);
-        assert_eq!(parsed.wal.write_buffer_size, 256 * 1024);
+        assert_eq!(parsed.wal.write_buffer_size, 2 * 1024 * 1024);
         assert_eq!(parsed.cluster_transport.raft_tick_interval_ms, 10);
         // New ClusterTransportTuning fields.
         assert_eq!(
--- a/nodedb-types/src/config/tuning/network.rs
+++ b/nodedb-types/src/config/tuning/network.rs
@@ -110,7 +110,7 @@ impl Default for WalTuning {
 }
 
 fn default_wal_write_buffer_size() -> usize {
-    256 * 1024
+    2 * 1024 * 1024
 }
 fn default_wal_alignment() -> usize {
     4096
--- a/nodedb-wal/src/writer.rs
+++ b/nodedb-wal/src/writer.rs
@@ -27,14 +27,14 @@ use crate::align::{AlignedBuf, DEFAULT_ALIGNMENT};
 use crate::error::{Result, WalError};
 use crate::record::{HEADER_SIZE, WalRecord};
 
-/// Default write buffer size: 256 KiB.
+/// Default write buffer size: 2 MiB.
 ///
 /// This is the batch size for group commit. Records accumulate here until
 /// the buffer is full or `sync()` is called.
 ///
 /// Matches `WalTuning::write_buffer_size` default. Override via
 /// `WalWriterConfig::write_buffer_size` at construction time.
-pub const DEFAULT_WRITE_BUFFER_SIZE: usize = 256 * 1024;
+pub const DEFAULT_WRITE_BUFFER_SIZE: usize = 2 * 1024 * 1024;
 
 /// Configuration for the WAL writer.
 #[derive(Debug, Clone)]
--- a/nodedb/src/config/server/env.rs
+++ b/nodedb/src/config/server/env.rs
@@ -338,6 +338,35 @@ pub fn apply_env_overrides(config: &mut ServerConfig) {
         apply_bool_env("NODEDB_TLS_ILP", &mut tls.ilp);
     }
 
+    // ── WAL tuning ─────────────────────────────────────────────────
+
+    if let Ok(val) = std::env::var("NODEDB_WAL_WRITE_BUFFER_SIZE") {
+        match parse_memory_size(&val) {
+            Ok(size) if size >= 64 * 1024 => {
+                tracing::info!(
+                    env_var = "NODEDB_WAL_WRITE_BUFFER_SIZE",
+                    value = size,
+                    "environment variable override applied"
+                );
+                config.tuning.wal.write_buffer_size = size;
+            }
+            Ok(size) => {
+                tracing::warn!(
+                    env_var = "NODEDB_WAL_WRITE_BUFFER_SIZE",
+                    value = size,
+                    "ignoring value below minimum 64KiB, using config value"
+                );
+            }
+            Err(_) => {
+                tracing::warn!(
+                    env_var = "NODEDB_WAL_WRITE_BUFFER_SIZE",
+                    value = %val,
+                    "ignoring malformed environment variable, using config value"
+                );
+            }
+        }
+    }
+
     // ── Observability overrides (PromQL, OTLP) ─────────────────────
 
     super::observability::apply_observability_env(&mut config.observability);
--- a/nodedb/src/control/server/ilp_listener.rs
+++ b/nodedb/src/control/server/ilp_listener.rs
@@ -73,7 +73,7 @@ impl IlpListener {
                                         Ok(tls_stream) => {
                                             let cs = ConnStream::tls(tls_stream);
                                             if let Err(e) = handle_ilp_connection(cs, peer, &state).await {
-                                                debug!(%peer, error = %e, "ILP TLS connection error");
+                                                warn!(%peer, error = %e, "ILP TLS connection error (data may be lost)");
                                             }
                                         }
                                         Err(e) => {
@@ -86,7 +86,7 @@ impl IlpListener {
                                 connections.spawn(async move {
                                     let cs = ConnStream::plain(stream);
                                     if let Err(e) = handle_ilp_connection(cs, peer, &state).await {
-                                        debug!(%peer, error = %e, "ILP connection error");
+                                        warn!(%peer, error = %e, "ILP connection error (data may be lost)");
                                     }
                                     drop(permit);
                                 });
--- a/nodedb/src/data/executor/handlers/timeseries.rs
+++ b/nodedb/src/data/executor/handlers/timeseries.rs
@@ -877,12 +877,19 @@ impl CoreLoop {
         format: &str,
         wal_lsn: Option<u64>,
     ) -> Response {
-        // LSN-based deduplication for WAL catch-up replays.
-        if let Some(lsn) = wal_lsn {
-            // Skip if already ingested into memtable.
-            if let Some(&max_lsn) = self.ts_max_ingested_lsn.get(collection)
-                && lsn <= max_lsn
-            {
+        // LSN-based deduplication: only skip records that are provably
+        // already flushed to sealed disk partitions. We do NOT track a
+        // max-ingested-LSN watermark because SPSC drops create gaps —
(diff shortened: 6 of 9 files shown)


emanzx:
## Verification v2 — commit f229ff7

### Issue #12 — Response frame chunking: FIXED

```sql
SELECT qname, COUNT(*) FROM dns_test GROUP BY qname
→ 553,217 rows in 2,167ms
```

Previously errored with `response frame too large: 24468546`. Now returns all 553K groups correctly. GROUP BY client_ip (10K) also works at 623ms.

### Issue #11 — WAL catch-up stall: STILL NOT FIXED

Same behavior as previous build. After 10M ingest:

```
Immediately: COUNT = 5,782,773 (58%) | Parts: 6 | WAL: 9
+5s  idle:   COUNT = 5,782,773          | Parts: 6 | WAL: 9
+30s idle:   COUNT = 5,782,773          | Parts: 6 | WAL: 9
+60s idle:   COUNT = 5,782,773          | Parts: 6 | WAL: 9
+120s idle:  COUNT = 5,782,773          | Parts: 6 | WAL: 9
```

Zero progress over 2 minutes. No new partitions, no WAL segments drained. The idle memtable flush that worked in the previous build (where we saw 5→6 partitions at +30s) didn't trigger this time either.

The WAL catch-up task only runs during active ingest. After ingest stops, 9 WAL segments (~4.2M rows) remain permanently invisible.

### Performance

| Metric | This build | Previous build |
|---|---|---|
| Ingest rate | 116K/s | 147K/s (slightly slower) |
| RSS peak | 185MB | 191MB |
| RSS idle | 100MB | 102MB |

### Summary

| Fix | Status |
|---|---|
| #12 Response frame chunking | ✅ **FIXED** — 553K rows returned successfully |
| #11 RSS regression | ✅ Fixed (100MB idle) |
| #11 WAL stall after ingest | ❌ **Still broken** — WAL segments never drain during idle |

emanzx:
## Verification v4 — commit a8a0836: ALL ISSUES FIXED

The 256KB WAL buffer was the root cause all along. Every wide ILP batch (280-580KB with 10+ columns) was silently dropped at `debug` level. We were chasing ghosts through the catch-up task, LSN watermarks, and backpressure gates — but the data never made it to WAL in the first place.

### Issue #11 — WAL visibility: FIXED

```
Ingest: 10,000,000 rows in 90.2s (110,904/s)
COUNT immediately after: 10,000,000 ← 100% VISIBILITY
COUNT after restart:     10,000,000 ← 100% CRASH RECOVERY
```

| Build | Visibility | Partitions |
|---|---|---|
| Original (pre-PR) | 931K / 10M (9%) | 1 |
| PR #10 (0a60295) | 2.1M / 10M (21%) | 2 |
| PR #13 v1 (2049f16) | 5.8M / 10M (58%) | 6 |
| PR #13 v3 (1ea2ba1) | 2.4M / 10M (24%) | 3 |
| **PR #13 v4 (a8a0836)** | **10M / 10M (100%)** | **20** |

### Issue #12 — GROUP BY frame size: FIXED

```sql
SELECT qname, COUNT(*) FROM dns_test GROUP BY qname
→ 1,996,464 rows in 7,188ms
```

Previously: `ERROR: response frame too large: 24468546`
Now: returns all 2M unique qnames without error.

### Full query battery at 10M (after restart)

```
COUNT(*)                           87ms (1 rows)
WHERE qtype='A' COUNT              719ms (correct filtering)
WHERE rcode='SERVFAIL' COUNT       249ms (correct filtering)
WHERE rcode='NONEXISTENT'          125ms (correct — empty result)
GROUP BY qtype                     739ms (8 rows)
GROUP BY client_ip (10K)           848ms (9,997 rows)
GROUP BY qname (2M unique)         7,188ms (1,996,464 rows)
time_bucket 1h                     163ms (3 buckets)
```

All queries return correct results across all 20 sealed partitions.

### Resources

```
Ingest rate:  110,904/s
RSS (ingest): 135MB steady
RSS (query):  4,974MB (loading all partitions for full scan)
Disk:         2,033MB
Partitions:   20
```

### Summary

Both #11 and #12 fully resolved. The journey:

1. We stress-tested with 10M high-cardinality rows (10K IPs, 50K+ qnames)
2. Found only 9-58% of data was queryable depending on the build
3. Investigated backpressure gates, LSN watermarks, idle flush timers, startup replay
4. Farhan found the actual root cause: **256KB WAL write buffer too small for wide ILP batches**, errors logged at `debug` level (invisible by default)
5. Fix: buffer 256KB → 2MB, errors `debug!` → `warn!`, removed unnecessary catch-up complexity

Ship it.

PR #93, opened 2026-05-02
Summary: Claude - Ensure can run in Mac

leowmjw:


leowmjw:
Wrong repo .. sorry! Use if it make sense; did not run in-depth analysis on the EventFd choice yet.  It runs on single node on mac; did not test comprehensively

PR #128, opened 2026-06-08
fix: gate Linux-only runtime paths

habibtalib:
## Summary
- gate Linux-only eventfd/io_uring and sd_notify paths so non-Linux builds can compile safely
- add non-Linux-safe fallbacks for readiness signalling and runtime bridge paths
- stabilize WAL mmap observability tests by serializing global counter assertions
- keep Linux-specific behavior covered by target-gated tests

## Verification
Ran on darkbuntu (Linux x86_64, rustc 1.96.0):
- cargo fmt --all -- --check
- cargo check --workspace --all-targets
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test -p nodedb-bridge --all-targets
- cargo test -p nodedb-wal --all-targets partially: lib/tests passed; bench target wal_throughput timed out under cargo test --all-targets on this host
- cargo test -p nodedb-cluster readiness

Note: full cargo test --workspace was not run due to the WAL benchmark timeout when cargo test executes bench binaries on the dev host.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2272,6 +2272,8 @@ checksum = "28a80e3145d8ad11ba0995949bbcf48b9df2be62772b3d351ef017dff6ecb853"
 [[package]]
 name = "fluxbench"
 version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "323b4f98b988b9779c6aa3f93781ef7421e36e73a5c610497857b2c01398acdc"
 dependencies = [
  "fluxbench-cli",
  "fluxbench-core",
@@ -2286,6 +2288,8 @@ dependencies = [
 [[package]]
 name = "fluxbench-cli"
 version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "42072e01658451938143f620a08f13b308e30a7fe41f3de87a79f351c5753328"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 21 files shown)


farhan-syah:
Pushed three commits to extend this PR from "compiles on non-Linux" to "runs on non-Linux (macOS)".

### Context
With the platform gating in this PR, the workspace compiles on macOS, but the server still aborts at startup:

```
Error: io error: eventfd-backed data-plane wake is only available on Linux
```

The data-plane core wake (`nodedb/src/data/eventfd.rs`) had no functional non-Linux backend — `EventFd::new()` returned `Unsupported`, which propagates through `spawn_core()` and fails data-plane bootstrap.

### Changes
- **`fix(runtime)`** — add a self-pipe `EventFd` fallback for non-Linux Unix targets (macOS, BSD), mirroring the pipe-based `EventFd` already used in `nodedb-bridge`. `notify()` writes a byte, `drain()` reads until empty, `poll_wait()` polls the read end; both ends are non-blocking + close-on-exec. Linux keeps the `eventfd` semaphore fast path unchanged; non-Unix targets (wasm, Windows) still return an explicit error.
- **`fix(cluster)`** — gate the `warn` import in `readiness.rs` to Linux. It is only used by the systemd path, so on non-Linux it was unused and broke `clippy -D warnings`.
- **`test(graph)`** — make the `CREATE GRAPH INDEX` dispatch-batching budget debug-aware (5s under `debug_assertions`, 1s for release). Debug builds run several times slower; the serial-loop regression scales by the same factor, so the guard still trips.

### Verification (macOS 15.6, aarch64)
- `cargo build` — builds
- Server boots, binds pgwire/native/HTTP listeners, and serves SQL end-to-end (verified with `psql`: `CREATE TABLE` / `INSERT` / `SELECT` / `count(*)`)
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- `cargo nextest run` — data-plane wake unit tests and the graph timing tests pass

### Note for non-Linux contributors
On a debug build (no io_uring, unoptimized), a few heavy graph integration tests can exceed nextest's default timeout purely from CPU oversubscription — they pass when run with reduced parallelism (e.g. `-j`) or under `--release`. This is environmental, not a correctness issue.

PR #176, opened 2026-07-10
feat(control): native SQL in-txn read-your-own-writes (U3) + cross-cluster txn_id groundwork

pizofreude:
## What & why

Verifies and closes **design unit U3 (in-transaction read-your-own-writes)** on the SQL paths, and lands the cluster-wire groundwork needed for the cross-node continuation (U4/U5).

Three focused commits:

1. **`fix(control): ensure native session on first-frame BEGIN`** — a `BEGIN` arriving on the very first native frame no longer races session creation, so the transaction has a session to attach to.
2. **`fix(control): propagate txn_id through the gateway for native in-txn RYOW`** — threads the session `txn_id` through the native gateway dispatch so an in-block read resolves this transaction's staging overlay. Single-node native in-txn RYOW now works.
3. **`feat(control): thread txn_id across cluster wire; defer cross-node RYOW to U4/U5`** — adds `ExecuteRequest.txn_id` (`WIRE_VERSION` 2→3, coordinated upgrade), threads the id through dispatcher/stream/receiver/all-cores, and mints globally-unique ids via `TxnId::from_origin(node_id, seq)`.

### Scope decision: U3 closed as single-node; cross-node deferred to U4/U5

Per the LOCKED design §8, U3's validation surface is **single-node only** ("single-node-verifiable: U1, U2, U3 (RYOW), U7"; cross-node is "cluster-bound (stage-2): U4, U5, U6, U8"). A 3-node probe (`cross_node_ryow_probe.rs`) is included but **`#[ignore]`d** as the executable spec for U4/U5.

**Confirmed root cause of the cross-node gap** (live trace): the per-transaction machinery — staging gate, COMMIT buffer, COMMIT/ROLLBACK overlay flush — is local-only. On a non-owner node the pgwire dispatcher takes the remote-leader forward short-circuit (`routing/execute.rs` `should_forward_via_gateway` → `dispatch_tasks_via_gateway`) and returns *before* `dispatch_task_loop`, so `route_in_tx_write` never runs. The in-txn write ships to the owner as a plain replicable write (Raft-proposed = committed at statement time, never staged), and `gateway_dispatch.rs` hardcodes `QueryContext.txn_id = None`, so the read arrives with no txn id. The wire `txn_id` field here is a necessary-but-not-sufficient prerequisite for the U4/U5 fix.

## How to test

```bash
# Single-node native in-txn RYOW — passes
cargo nextest run -p nodedb-cluster-tests -E 'test(native_sql_ryow)'

# Cross-node probe — ignored by default; run explicitly to reproduce the U4/U5 gap
cargo nextest run -p nodedb-cluster-tests -E 'test(cross_node_in_txn_select_reads_own_write)' --run-ignored all
```

Also verified locally: `cargo fmt --all` clean; `cargo check -p nodedb` clean (zero warnings); `cargo clippy` introduces **no new warnings** in any file this branch touches.

## Tradeoffs / alternatives considered

- **Coordinated (not rolling) wire upgrade** for `WIRE_VERSION` 2→3 — the field is additive but mixed-version clusters are out of scope for this branch.
- **Defer cross-node RYOW** rather than build the cross-shard-transaction machinery (gateway-routed staging / COMMIT / ROLLBACK / overlay-drop) ahead of the U4/U5 schedule and out of order.
- **`TxnId::from_origin(node_id, seq)`** kept as groundwork so ids are globally unique before cross-node txn coordination exists (all non-txn call sites pass `None`/identity).

## Notes for review

- Touches wire dispatch on pgwire/native/RESP paths and bumps `WIRE_VERSION` — flagging per the "wire protocol changes" note in `CONTRIBUTING.md`.
- CI is opt-in via the `run-ci` label. Heads-up: the workspace currently has pre-existing clippy lints under current stable (`iter_kv_map`, `question_mark`, `unneeded_wildcard_pattern`) in files **unrelated to this branch**; with `-D warnings` they would trip the lint job independently of this change.

### DIFF
--- a/nodedb-cluster-tests/tests/cluster_execute_request.rs
+++ b/nodedb-cluster-tests/tests/cluster_execute_request.rs
@@ -53,6 +53,7 @@ fn make_kv_put_request(
             collection: collection.into(),
             version: descriptor_version,
         }],
+        txn_id: None,
     }
 }
 
@@ -209,6 +210,7 @@ async fn execute_request_cross_node_dispatch() {
             collection: "cross_node_kv".into(),
             version: 0, // Accept any version (pre-B.1 sentinel bypass)
         }],
+        txn_id: None,
     };
 
     let resp = send_execute_request(sender_transport, target_addr, req).await;
--- a/nodedb-cluster-tests/tests/cross_node_ryow_probe.rs
+++ b/nodedb-cluster-tests/tests/cross_node_ryow_probe.rs
@@ -1,0 +1,145 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+//! DEFERRED PROBE (design unit U4/U5) — CROSS-NODE in-transaction
+//! read-your-own-writes. `#[ignore]`d: cross-node in-txn RYOW is out of scope
+//! for U3, whose validation surface is single-node only (LOCKED design §8:
+//! "single-node-verifiable: U1, U2, U3 (RYOW), U7"; cross-node is "cluster-bound
+//! (stage-2): U4, U5, U6, U8"). Kept in-tree, ignored, as the executable
+//! specification for the U4/U5 cross-shard-transaction work.
+//!
+//! SINGLE-NODE in-txn RYOW works today (U3): an in-block read takes the
+//! NON-streaming dispatch path (`maybe_stream_select` bails on `InBlock`), and
+//! `dispatch_local` threads the session `txn_id` to the local leaseholder so the
+//! Data Plane merges this transaction's staging overlay. The native twin
+//! (`native_sql_ryow_probe.rs`) and the graph twin
+//! (`single_node_calvin_graph_txn.rs`) both cover that and stay GREEN.
+//!
+//! CONFIRMED ROOT CAUSE of the cross-node gap (diagnosed 2026-07, live trace):
+//! the whole per-transaction machinery — staging gate, COMMIT buffer, and the
+//! COMMIT/ROLLBACK overlay flush — is LOCAL-ONLY. On a NON-owner node the pgwire
+//! dispatcher takes the remote-leader forward short-circuit
+//! (`routing/execute.rs`: `should_forward_via_gateway` → `dispatch_tasks_via_gateway`)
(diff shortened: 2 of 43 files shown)


farhan-syah:
Single-node native RYOW looks right. Two asks before merge:

1. **Drop the internal roadmap labels from source.** The new test files/comments carry `U1`/`U4/U5`, "design §8", "GAP B", etc. — describe the behavior instead (e.g. `cross_node_ryow_probe.rs` → name/ignore-reason about "non-owner forwards write without threading the session overlay"). The technical notes are useful; just the unit labels need to go.

2. **The `WIRE_VERSION` 2→3 bump has no consumer yet.** `txn_id` is stamped `None` everywhere and cross-node RYOW is deferred, so this forces a coordinated cluster upgrade for an inert field. Suggest folding the wire break into the change that actually reads it — one upgrade at feature time instead of one now for groundwork.

Commits 1–2 are fine to land independently if you want to split the wire groundwork out.

pizofreude:
Thanks for the review, Farhan.

Closing this PR as obsolete against current `main`:

1. **Wire groundwork (commit 3)** is superseded. `feat(cluster): forward transaction id on ExecuteRequest RPC` already landed `ExecuteRequest.txn_id` without a `WIRE_VERSION` bump, which matches your guidance to fold wire breaks into the change that actually consumes them. This branch is also ~160+ commits behind and not worth rebasing that groundwork.
2. **Internal roadmap labels** in the probe/comments would need stripping per your ask and the updated CONTRIBUTING rules. Easier to start clean from tip.
3. **Intent next:** open a new tip-based PR that only fixes (a) native first-frame BEGIN creating a session, and (b) threading the session `txn_id` through the native gateway **local** dispatch path for in-txn SQL read-your-own-writes — behavior-named tests only, no wire version bump, no internal unit labels.

Appreciate the clear review criteria; applying them going forward.

pizofreude:
Obsolete vs tip (ExecuteRequest.txn_id already landed; branch too far behind). Follow-up will be a slim native gateway txn-overlay PR from current main.

PR #196, opened 2026-07-21
fix(native): dispatch txn commit batch to the owning vshard

### LINKED ISSUE #193: Native protocol: a row committed inside BEGIN/COMMIT is invisible to PK point lookups and filtered COUNT(*) — full scans see it

`origin/main @ eea86b279`

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Document (strict)

### Summary

Over the native binary protocol (`:6433`), a row inserted inside an explicit transaction (BEGIN op → SQL INSERT → COMMIT op) is durably stored — full scans return it and it survives reconnects — but a PK point lookup (`WHERE id = '<pk>'`) returns 0 rows, on the same connection and on a fresh one, and a filtered `count(*)` misses it too. The identical statement sequence over pgwire (`:6432`) is correct, and an autocommit insert over native is also correct — the trigger is specifically the transactional write path on the native transport. Looks adjacent to the session plan cache fixed in v0.4.0 for pgwire point reads (that fix did not help the native path).

### Steps to reproduce

Native-protocol session against `:6433` (requests are the documented MessagePack frames `[op, seq, [0x01, {field => value}]]`; `0x20` = SQL, `0x40`/`0x41` = BEGIN/COMMIT ops). SQL payloads and returned rows:

```sql
-- fresh data directory; authenticated native session
CREATE COLLECTION n_probe (id TEXT PRIMARY KEY, name TEXT) WITH (engine='document_strict');
-- status OK

-- op BEGIN (0x40)
INSERT INTO n_probe (id, name) VALUES ('a1', 'alpha');   -- rows_affected 1
-- op COMMIT (0x41)

SELECT id FROM n_probe WHERE id = 'a1';                  -- (0 rows)   <- WRONG
SELECT id FROM n_probe;                                  -- a1 (1 row) -- durably stored
SELECT count(*) FROM n_probe WHERE name = 'alpha';       -- 0          <- WRONG

-- fresh native connection:
SELECT id FROM n_probe WHERE id = 'a1';                  -- (0 rows)   <- WRONG, persists across sessions

-- control, same session, autocommit (no BEGIN/COMMIT ops):
INSERT INTO n_probe (id, name) VALUES ('b2', 'beta');
SELECT id FROM n_probe WHERE id = 'b2';                  -- b2 (1 row) -- correct
```

The same SQL sequence over pgwire on `:6432` (psql, `BEGIN;`/`COMMIT;` as statements) returns `a1` from the point lookup and `1` from the filtered count.

### Expected behavior

Committed transactional writes are visible to every read path on the native transport — point lookups and filtered aggregates — identical to full scans, to autocommit writes, and to pgwire.

### Actual behavior

PK point lookups and filtered `count(*)` permanently return 0 rows/0 for rows written through BEGIN/COMMIT over native, on the writing connection and all later ones, while full scans return the row. No error anywhere.

- [x] Core functionality is broken with no acceptable workaround *(on the native transport; pgwire unaffected)*

### Proposed severity

SEV-2 — High: silently-wrong reads of durably stored data; the stored data itself is intact (scans prove it). "Workaround" is switching transports or avoiding transactions, not a rewrite within the native path.

### Reproducibility

Always — every attempt

_(blank — first observed on `3eaa49873`; unknown whether older builds were affected)_

### Environment & logs

Linux x86_64, release build. No server-side errors or warnings logged during the sequence. Can attach a packet capture of the native session if useful.

### Before submitting

laksamanakeris:
Fixes #193.

## Problem

A row committed inside an explicit transaction over the native binary protocol (`Begin` op → SQL INSERT → `Commit` op) was durably stored but permanently invisible to PK point lookups and filtered `count(*)` — full scans saw it, and the same SQL over pgwire was correct.

Root cause (full trace in [the issue comment](https://github.com/NodeDB-Lab/nodedb/issues/193#issuecomment-5029817822)): both transports share `run_commit` → `dispatch_single_shard`, which builds the commit tasks (`MetaOp::ResolveTxn` + `MetaOp::TransactionBatch`) with the correct `vshard_id`. The pgwire seam submits them with that `vshard_id`; the native seam (`NativeTxnDp::dispatch_no_wal`) sent them through `gw.execute(...)`, which drops `task.vshard_id` and re-routes from the plan. `MetaOp` plans have no collection name, so the router's `primary_vshard` fell back to vShard 0 — the whole commit batch (row write + index maintenance) was applied on vShard 0's core instead of the collection's owning vShard.

## Fix

- `NativeTxnDp::dispatch_no_wal` now always dispatches through the direct write path with `task.vshard_id`, matching pgwire's `dispatch_task_no_wal` and native's own pre-gateway fallback branch (which was already correct).
- Guardrail: `route_plan` rejects `ResolveTxn` / `TransactionBatch` plans with an error instead of silently misrouting them to vShard 0, so any future caller that sends a commit meta-op through the gateway fails loudly instead of corrupting placement.

## Tests

- New integration test `native_txn_commit_visibility.rs`: the issue's exact repro (Begin op / INSERT / Commit op over the native wire), asserting the committed row is visible to a PK point lookup and a filtered `count(*)` on the writing connection and on a fresh connection, with the full scan as control. The test wires the gateway the way production boot does and runs 4 data-plane cores with a collection homed off core 0 — reproducing the bug requires both (the default single-core, gateway-less test harness masks it; verified red before the fix, green after).
- New router unit test `commit_meta_ops_are_rejected` for the guardrail.
- Full `cargo test -p nodedb` suite passes (one pre-existing, unrelated failure on `main`: `shuffle_grace_infeasible_budget_is_deterministic_error`, fails identically on `main @ 48281658`).

## Notes for reviewers

- Rows already committed through native-protocol transactions on affected multi-core deployments remain mis-homed on vShard 0 after this fix (scan-visible, lookup-invisible). Flagged in the changelog; a repair/re-home tool likely deserves a separate issue.
- The multi-shard commit branch is unaffected: it flushes through the Calvin sequencer, not this seam, and every remaining `dispatch_no_wal` call site goes through the fixed seam.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,6 @@ NodeDB uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ---
 
-## [Unreleased]
-
-### Fixed
-
-- **Native protocol transactions** — a row committed inside an explicit transaction over the native binary protocol (`Begin` op → SQL DML → `Commit` op) is now visible to PK point lookups and filtered aggregates, not just full scans (#193). The native COMMIT seam routed the commit's `ResolveTxn`/`TransactionBatch` meta-ops through the gateway, which cannot derive a route for collection-less meta plans and fell back to vShard 0 — durably applying the commit batch on the wrong core. Native COMMIT now dispatches with the task's pre-classified vShard, matching pgwire, and the gateway router rejects commit meta-ops outright instead of silently misrouting them. **Note:** rows committed through native-protocol transactions on affected multi-core builds were applied to vShard 0's core and remain there after upgrading — they stay visible to full scans but invisible to point lookups; re-inserting them (or dump/reload) re-homes them.
-
----
-
 ## [0.4.0] - 2026-07-20
 
 NodeDB 0.4.0 is a substantial distributed-correctness and durability release. It adds cross-shard transactional execution and distributed query/graph processing, extends temporal and sparse-vector SQL, and hardens replication, recovery, indexing, authentication, and sync across the storage engines.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
farhan-syah, line 1: I would suggest, fold it in 0.4.0, as changes from 0.3.0 to 0.4.0 (net changes), and keep it short, we don't want to make the changelog too long.

farhan-syah (changes requested):

CHANGELOG.md
laksamanakeris: Done in 754a1f6c — dropped the Unreleased section and folded it into 0.4.0 Fixed as a single short bullet.

farhan-syah (approved):

PR #206, opened 2026-07-22
fix: join LIMIT applied before WHERE, and same-named output columns collapsing

### LINKED ISSUE #200: LIMIT on a join between virtual catalog tables returns 0 rows

`origin/main @ 74febcf80`

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Not engine-specific / unsure

### Summary

A query joining two virtual pg_catalog tables returns the correct rows — until a `LIMIT` clause is added, at which point the same query silently returns 0 rows. Real (user) collections joined with LIMIT are unaffected; the failure is specific to the virtual catalog evaluator. Any catalog introspection that pages with LIMIT silently sees an empty catalog.

### Steps to reproduce

```sql
CREATE COLLECTION wk_probe (id TEXT PRIMARY KEY, name TEXT DEFAULT 'x', qty INT) WITH (engine='document_strict');

SELECT t.oid, t.typname FROM pg_type as t LEFT JOIN pg_range as r ON oid = rngtypid
 WHERE t.typname IN ('int2','text');
--  21 | int2
--  25 | text          (2 rows — correct)

SELECT t.oid, t.typname FROM pg_type as t LEFT JOIN pg_range as r ON oid = rngtypid
 WHERE t.typname IN ('int2','text') LIMIT 5;
-- (0 rows)            <- WRONG: LIMIT 5 on a 2-row result must return both rows

SELECT a.attname FROM pg_attribute a
  LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum
 WHERE a.attrelid = 'wk_probe'::regclass LIMIT 5;
-- (0 rows)            <- same shape, different catalog pair; without LIMIT: id/name/qty

-- control: real collections, join + LIMIT — correct
CREATE COLLECTION wk_b (id TEXT PRIMARY KEY, ref TEXT) WITH (engine='document_strict');
INSERT INTO wk_probe (id, name, qty) VALUES ('a','n',1);
INSERT INTO wk_b (id, ref) VALUES ('b1','a');
SELECT w.id AS w_id, b.id AS b_id FROM wk_probe w LEFT JOIN wk_b b ON b.ref = w.id LIMIT 5;
--  a | b1            (1 row — correct)
```

### Expected behavior

`LIMIT n` truncates a result set to at most n rows; it never empties a non-empty result. Catalog joins behave like collection joins.

### Actual behavior

Catalog-to-catalog joins return 0 rows whenever a LIMIT clause is present; the identical query without LIMIT returns the correct rows. Single-table catalog queries with LIMIT are fine, and collection joins with LIMIT are fine.

### Proposed severity

### Reproducibility

Always — every attempt

_(blank — catalog joins only recently became usable, so this shape may never have worked with LIMIT)_

### Environment & logs

Linux x86_64, release build, fresh data directory. Statements succeed silently — no server log lines. Best guess: the virtual-catalog evaluator applies LIMIT before (or instead of) materializing the join output.

### Before submitting

### LINKED ISSUE #201: Qualified same-name columns in a join select list all resolve to the last table's value

`origin/main @ 74febcf80`

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Document (strict)

### Summary

When two joined collections share a column name (both have `id`), projecting the column qualified from each table — `SELECT w.id, b.id` — returns the last-joined table's value in BOTH output columns. Attaching aliases (`w.id AS w_id, b.id AS b_id`) resolves correctly, and the join predicate itself evaluates correctly, so storage and joining are fine; only unaliased qualified projection is broken. Silently wrong results, no error.

### Steps to reproduce

```sql
CREATE COLLECTION wk_probe (id TEXT PRIMARY KEY, name TEXT) WITH (engine='document_strict');
CREATE COLLECTION wk_b (id TEXT PRIMARY KEY, ref TEXT) WITH (engine='document_strict');
INSERT INTO wk_probe (id, name) VALUES ('a','n');
INSERT INTO wk_b (id, ref) VALUES ('b1','a');

SELECT w.id, b.id, w.name, b.ref FROM wk_probe w JOIN wk_b b ON b.ref = w.id;
--  id | id | name | ref
--  b1 | b1 | n    | a        <- WRONG: first column must be 'a' (w.id)

SELECT w.id AS w_id, b.id AS b_id FROM wk_probe w LEFT JOIN wk_b b ON b.ref = w.id;
--  w_id | b_id
--  a    | b1                 (correct with aliases)
```

Note `w.name` and `b.ref` (unique names) project correctly in the same statement — only the colliding name is misresolved.

### Expected behavior

`w.id` projects wk_probe's value regardless of aliasing; table qualification alone disambiguates same-named columns, as in PostgreSQL.

### Actual behavior

Every unaliased projection of a column name shared by joined tables yields the same value — the last table's — while aliased projections of the identical expressions are correct.

### Proposed severity

### Reproducibility

Always — every attempt

_(blank — unknown)_

### Environment & logs

Linux x86_64, release build, fresh data directory. No errors emitted. Best guess: select-list resolution keys output columns by bare column name after qualification is dropped, so the last binding for `id` wins for all occurrences.

### Before submitting

mkhairi:
Closes #200
Closes #201

## What this changes and why

Two independent SELECT correctness bugs, one commit each, both with reproducing tests.

### 1. `fix(sql)`: join LIMIT was applied before post-join WHERE filters

`execute_hash_join` passed the user `LIMIT n` straight into the probe, so the probe emitted the first `n` ON-matched rows and `filter_and_project` evaluated the WHERE predicate on that already-truncated set. Rows satisfying the WHERE clause that sorted after the first `n` probe matches were dropped — usually emptying the result.

Real-collection joins push their WHERE into the scanned side, leaving `post_filters` empty, so the ordering was harmless there. Virtual catalog joins keep both sides as inline `ProviderScan`s and carry the WHERE as a join post-filter, which is where it bit:

```sql
-- returns 1 row
SELECT t.typname FROM pg_type t
  LEFT JOIN pg_range r ON t.oid = r.rngtypid
  WHERE t.typname = 'int4';

-- returns 0 rows (expected: the same 1 row)
SELECT t.typname FROM pg_type t
  LEFT JOIN pg_range r ON t.oid = r.rngtypid
  WHERE t.typname = 'int4' LIMIT 5;
```

The probe now runs under the byte-budget ceiling whenever post-filters are present, and the user LIMIT is applied after filtering. A narrow LIMIT still truncates; a wide one no longer under-fills. The no-post-filter path is untouched, so the existing `ResourcesExhausted` budget semantics are unchanged.

### 2. `fix(pgwire)`: two output columns sharing a name collapsed to one value

```sql
CREATE COLLECTION w (id TEXT PRIMARY KEY, wname TEXT) WITH (engine='document_strict');
CREATE COLLECTION b (id TEXT PRIMARY KEY, w_id TEXT) WITH (engine='document_strict');
INSERT INTO w (id, wname) VALUES ('w1','alpha');
INSERT INTO b (id, w_id) VALUES ('b1','w1');

SELECT w.id, b.id FROM w JOIN b ON b.w_id = w.id;
--  b1 | b1     (expected: w1 | b1)

SELECT w.id AS w_id, b.id AS b_id FROM w JOIN b ON b.w_id = w.id;
--  w1 | b1     (correct)
```

The planner and join executor were both correct — the merged row carries `w.id` and `b.id`, and `project_row` reads each by its qualified lookup key. The loss happened at the last step: a shaped row is a `Map<String, JsonValue>` keyed by the column's *display* name, and an unaliased qualified projection displays as its bare last segment. Both columns keyed on `id`, so the second insert overwrote the first, and every encoder then re-read that single cell by name. Aliasing produced distinct display names, which is why only the unaliased spelling was wrong.

Shaped rows now store each column's cell under a unique key from `response_shape::project::cell_keys` — the display name for the first occurrence, `<name>_<n>` for later duplicates, skipping any candidate that would shadow another column's real name. Every consumer reads cells through `ShapedRows::cell_keys()` so they stay aligned: the pgwire non-streaming and streaming encoders, the native converter, and the HTTP serializers. Because HTTP writes a row map directly to JSON, the suffix is user-visible there — a duplicate-name projection emits `{"id": …, "id_1": …}`, since a JSON object cannot repeat a key either and dropping the duplicate would lose a projected column. pgwire and native are positional on the wire and still report both columns as `id`. For a duplicate-free column list the mapping is the identity, so nothing else changes; wire column names are untouched (duplicates still describe as `id`, `id`, matching PostgreSQL).

## How to test it

```bash
cargo nextest run -p nodedb --test pg_catalog_reflection --test sql_join_correctness
cargo nextest run -p nodedb --lib -E 'test(cell_keys) or test(project_row)'
```

New coverage:

- `pg_catalog_reflection`: catalog join + WHERE + wide LIMIT (returns the matching rows) and + narrow LIMIT (still truncates).
- `sql_join_correctness`: real-collection analogues of both, plus unaliased and aliased qualified same-name projections.
- unit tests for `cell_keys` (identity without duplicates, `_n` suffixing, shadow-avoidance) and duplicate-name `project_row`.

Verified as genuine regression guards: with the two source changes reverted and the tests kept, exactly the two bug repros fail and the rest stay green.

## Tradeoffs / alternatives considered

- **LIMIT fix**: pushing the post-filter into the probe predicate would let the probe stop early and stay O(n). That is a larger change to `probe_hash_index`'s filter handling; this PR takes the minimal correct ordering fix. Cost is that a post-filtered join with a small LIMIT now probes up to the budget ceiling instead of stopping at `n` — correctness over an early exit that was producing wrong answers.
- **Duplicate-name fix**: the structurally cleaner fix is a positional row representation (`Vec<JsonValue>` instead of a name-keyed `Map`), which removes the failure mode by construction. That touches every `ShapedRows` producer and consumer in the tree; the `cell_keys` indirection is contained and keeps that refactor open as a follow-up.

## Verified against the reported reproductions

Both issues' exact reproduction scripts were re-run against this branch:

**#200** — catalog join + `LIMIT`, both reported shapes:

```
SELECT t.oid, t.typname FROM pg_type as t LEFT JOIN pg_range as r ON oid = rngtypid
 WHERE t.typname IN ('int2','text') LIMIT 5;
--  21 | int2
--  25 | text          (was 0 rows)

SELECT a.attname FROM pg_attribute a
  LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum
 WHERE a.attrelid = 'wk_probe'::regclass LIMIT 5;
--  id / name / qty    (was 0 rows)
```

Each matches its no-LIMIT result exactly.

**#201** — unaliased qualified same-name projection:

```
SELECT w.id, b.id, w.name, b.ref FROM wk_probe w JOIN wk_b b ON b.ref = w.id;
--  a | b1 | n | a     (was b1 | b1 | n | a)

SELECT w.id AS w_id, b.id AS b_id FROM wk_probe w LEFT JOIN wk_b b ON b.ref = w.id;
--  a | b1             (unchanged, still correct)
```

## Test results

- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- `cargo nextest run -p nodedb --lib` — 4627/4628 pass
- integration sweep (catalog reflection/regclass/oid-stability/select-semantics, join correctness, subquery, lateral, recursive CTE, kv select, http + native result projection) — 81/81 pass

The one lib failure is `control::catalog_entry::tests::collection::purge_collection_is_scoped_to_database`, which fails identically on unmodified `bece8812d` (verified in a clean worktree) and is untouched by this PR.

### DIFF
--- a/nodedb/src/control/server/response_shape/project.rs
+++ b/nodedb/src/control/server/response_shape/project.rs
@@ -55,3 +55,69 @@ pub fn is_scan_wrapper(map: &serde_json::Map<String, serde_json::Value>) -> bool
         && matches!(map.get("id"), Some(serde_json::Value::String(_)))
         && matches!(map.get("data"), Some(serde_json::Value::Object(_)))
 }
+
+/// Unique per-column cell keys for a shaped column list.
+///
+/// SQL output column names may legally repeat — `SELECT w.id, b.id` displays
+/// both columns as `id` — but a shaped row is a JSON map and cannot hold two
+/// cells under one key: inserting the second cell would overwrite the first,
+/// making both wire columns render the last value. The shaper therefore
+/// stores the first occurrence of a name under the name itself and each later
+/// duplicate under `<name>_<n>` (n = 1, 2, …), skipping any candidate that
+/// collides with another column's name. Row writers and every protocol
+/// encoder derive keys through this one function so they always agree; for a
+/// duplicate-free column list this is the identity mapping.
+pub fn cell_keys(columns: &[String]) -> Vec<String> {
+    use std::collections::HashSet;
+
+    let names: HashSet<&str> = columns.iter().map(String::as_str).collect();
+    let mut used: HashSet<String> = HashSet::with_capacity(columns.len());
+    columns
+        .iter()
+        .map(|name| {
+            if used.insert(name.clone()) {
+                return name.clone();
+            }
+            let mut n = 1usize;
+            loop {
+                let candidate = format!("{name}_{n}");
+                if !names.contains(candidate.as_str()) && used.insert(candidate.clone()) {
+                    return candidate;
+                }
+                n += 1;
+            }
+        })
+        .collect()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::cell_keys;
+
+    fn keys(cols: &[&str]) -> Vec<String> {
+        cell_keys(&cols.iter().map(|s| s.to_string()).collect::<Vec<_>>())
+    }
+
+    /// Duplicate-free column lists map to themselves.
+    #[test]
+    fn cell_keys_is_identity_without_duplicates() {
+        assert_eq!(keys(&["id", "name", "score"]), ["id", "name", "score"]);
+        assert_eq!(keys(&[]), Vec::<String>::new());
+    }
+
+    /// Later duplicates get a `_n` suffix; the first occurrence keeps the
+    /// bare name (`SELECT w.id, b.id` → `id`, `id_1`).
+    #[test]
+    fn cell_keys_suffixes_later_duplicates() {
+        assert_eq!(keys(&["id", "id"]), ["id", "id_1"]);
+        assert_eq!(keys(&["id", "id", "id"]), ["id", "id_1", "id_2"]);
+    }
+
+    /// A suffix candidate never collides with a real column of that name:
+    /// with columns `id, id, id_1` the second `id` skips `id_1` (taken by a
+    /// real column) and becomes `id_2`.
+    #[test]
+    fn cell_keys_skips_candidates_that_shadow_real_columns() {
+        assert_eq!(keys(&["id", "id", "id_1"]), ["id", "id_2", "id_1"]);
+    }
+}
--- a/nodedb/src/control/server/http/routes/query_stream.rs
+++ b/nodedb/src/control/server/http/routes/query_stream.rs
@@ -122,6 +122,9 @@ pub(super) fn ndjson_body_stream(
                     return;
                 }
             };
+            // Row maps are keyed by `ShapedRows::cell_keys`, so each NDJSON
+            // line serializes as-is; two output columns sharing a name emit
+            // `{"id": …, "id_1": …}` rather than collapsing to one cell.
             let shaped = shape_decoded_rows(&value, projection.as_ref());
             for row in shaped.rows {
                 if emitted >= limit {
--- a/nodedb/src/control/server/http/routes/result_shape.rs
+++ b/nodedb/src/control/server/http/routes/result_shape.rs
@@ -47,6 +47,11 @@ pub(super) fn shape_http_payload(
         database_id,
         tenant_id,
     )? {
+        // Each row map is already keyed by `ShapedRows::cell_keys`, so it
+        // serializes to JSON as-is. When two output columns share a name the
+        // later one carries a `_<n>` suffix (`SELECT w.id, b.id` →
+        // `{"id": …, "id_1": …}`) — a JSON object cannot repeat a key, and
+        // dropping the duplicate would silently lose a projected column.
         ShapeOutcome::Rows(shaped) => Ok(HttpShaped::Rows(
             shaped
                 .rows
@@ -111,6 +116,8 @@ pub(super) fn ddl_results_to_json(
                     "tag": command,
                 }));
             }
+            // Keyed by `ShapedRows::cell_keys`, same JSON contract as
+            // `shape_http_payload` above (duplicate names take a `_<n>` key).
             DdlResult::Rows(shaped) => {
                 for row in shaped.rows {
                     rows.push(serde_json::Value::Object(row));
--- a/nodedb/src/control/server/native/dispatch/conversion.rs
+++ b/nodedb/src/control/server/native/dispatch/conversion.rs
@@ -146,15 +146,18 @@ pub(crate) fn calvin_native_response(
 /// `json_to_value_display`; a column absent from a given row's map becomes
 /// `Value::Null`.
 pub(crate) fn to_native_columns_rows(shaped: &ShapedRows) -> (Vec<String>, Vec<Vec<Value>>) {
+    // Cells live in the row maps under unique per-column keys (display names
+    // may repeat across columns, e.g. `SELECT w.id, b.id`); derive the same
+    // keys the shaper used so every column reads its own cell.
+    let cell_keys = crate::control::server::response_shape::project::cell_keys(&shaped.columns);
     let rows = shaped
         .rows
         .iter()
         .map(|row| {
-            shaped
-                .columns
+            cell_keys
                 .iter()
-                .map(|col| {
-                    row.get(col.as_str())
+                .map(|key| {
+                    row.get(key.as_str())
                         .map(json_to_value_display)
                         .unwrap_or(Value::Null)
                 })
--- a/nodedb/src/control/server/native/dispatch/conversion.rs
+++ b/nodedb/src/control/server/native/dispatch/conversion.rs
@@ -146,10 +146,10 @@ pub(crate) fn calvin_native_response(
 /// `json_to_value_display`; a column absent from a given row's map becomes
 /// `Value::Null`.
 pub(crate) fn to_native_columns_rows(shaped: &ShapedRows) -> (Vec<String>, Vec<Vec<Value>>) {
-    // Cells live in the row maps under unique per-column keys (display names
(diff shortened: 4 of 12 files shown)


nodedb/src/control/server/response_shape/project.rs
```diff
@@ -55,3 +55,69 @@ pub fn is_scan_wrapper(map: &serde_json::Map<String, serde_json::Value>) -> bool
...
+/// SQL output column names may legally repeat — `SELECT w.id, b.id` displays
+/// both columns as `id` — but a shaped row is a JSON map and cannot hold two
+/// cells under one key: inserting the second cell would overwrite the first,
+/// making both wire columns render the last value. The shaper therefore
+/// stores the first occurrence of a name under the name itself and each later
+/// duplicate under `<name>_<n>` (n = 1, 2, …), skipping any candidate that
+/// collides with another column's name. Row writers and every protocol
+/// encoder derive keys through this one function so they always agree; for a
+/// duplicate-free column list this is the identity mapping.
+pub fn cell_keys(columns: &[String]) -> Vec<String> {
```
farhan-syah, line 70: The `cell_keys` invariant ("read cells via these keys, never the raw display name") is enforced only by convention, and two consumers already bypass it: the HTTP paths dump the raw row map without re-keying —

- `http/routes/query_stream.rs:130` → `serde_json::Value::Object(row)`
- `http/routes/result_shape.rs:116` → same

So an HTTP/NDJSON `SELECT w.id, b.id` now emits `{"id":"w1","id_1":"b1"}`. That is strictly better than the old collapse-to-one-value, but the synthetic `id_1` is now a **user-visible, untested** wire contract on a consumer the PR description does not mention ("all three consumers" omits HTTP). A future change to the `_n` scheme would silently alter HTTP output.

Suggest computing the keys once and storing them on `ShapedRows` (e.g. a `cell_keys: Vec<String>` field) so every consumer — pgwire, native, **and** HTTP — shares one source of truth instead of re-deriving. That closes the HTTP gap, removes the redundant per-consumer recomputation, and keeps the deferred positional-`Vec` refactor open. At minimum, add an HTTP duplicate-name test and document that `id_1` is the intended JSON shape.

mkhairi:
Good catch on the HTTP gap — you're right that "all three consumers" was wrong, and that `id_1` was reaching clients as an undocumented, untested contract. Fixed in b04047b13.

`ShapedRows::cell_keys()` is now the single source of truth. pgwire and native read through it instead of each re-deriving, the `rows` field documents that cells are keyed by it rather than by display name, and all three HTTP sites (`shape_http_payload`, `ddl_results_to_json`, and the NDJSON stream) document the JSON shape: a duplicate-name projection emits `{"id": …, "id_1": …}`, because a JSON object cannot repeat a key either and dropping the duplicate would silently lose a projected column. pgwire and native stay positional and still report both columns as `id`, matching PostgreSQL.

`http_result_projection` now asserts both cells survive over **both** `/v1/query` and `/v1/query/stream`. The PR description has been corrected to name HTTP as a consumer and state the `id_1` contract.

One deviation from your suggestion: I used an accessor rather than a stored `cell_keys: Vec<String>` field. `ShapedRows` has 146 struct-literal construction sites, so a required field would pull that many unrelated edits into a two-bug fix PR. Derivation is per response, not per row, so the caching a field would buy isn't measurable. Happy to switch to the field if you'd prefer it — either as part of this PR or alongside the positional-`Vec` refactor, where all those literals have to be touched anyway.

Verification: `cargo fmt --all --check` clean, `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean, 49/49 across the catalog, join, native, and HTTP test files. I also timed `query_stream_returns_all_rows_across_chunks` since it runs ~57s and this change touches that path — it is 58.6s on the unmodified base, so that is pre-existing, not a regression from here.

Issue #223, opened 2026-07-26
Strict-schema engines (document_strict, kv) reject standard integer type spellings INT4 / INT8 / SMALLINT / INT2 at DDL time


emanzx:
### Version / build tested against

`origin/main @ f7fcc7718` (release build, fresh data dir; reproduced live today).

### Deployment mode

Origin — single node (local), via psql over pgwire.

### Engine(s) involved

`document_strict` and `kv` — the `ColumnType`-backed strict-schema engines. (Schemaless/columnar/timeseries engines accept these spellings — see #217 for their separate wire-fidelity defect.)

### Summary

The strict-schema engines reject the standard Postgres integer type spellings `INT4`, `INT8`, `SMALLINT`, and `INT2` at DDL time with `unknown column type`, while accepting `INT`, `INTEGER`, and `BIGINT`. The same spellings are accepted by the schemaless engine family, so the DDL surface is inconsistent across engines, and Postgres-compatible tooling that emits `int4`/`int2` spellings fails on strict engines only. Root cause: `ColumnType::from_str` (`nodedb-types/src/columnar/column_parse.rs:124`) matches only `"BIGINT" | "INT64" | "INTEGER" | "INT"`.

### Steps to reproduce

```sql
CREATE COLLECTION strict_probe (id TEXT PRIMARY KEY, a INT4) WITH (engine='document_strict');
-- ERROR:  parse error: unknown column type: 'INT4'

CREATE COLLECTION strict_probe2 (id TEXT PRIMARY KEY, f SMALLINT) WITH (engine='kv');
-- ERROR:  parse error: column 'f': unknown column type: 'SMALLINT'
```

### Expected behavior

`INT4`/`INT8`/`SMALLINT`/`INT2` accepted as aliases of the existing integer type (as the schemaless engines already do), so DDL is consistent across engines and standard Postgres type spellings work everywhere.

### Actual behavior

DDL fails with `parse error: unknown column type` on strict engines only.

### What actually happened? (check all that are true)

- [ ] Acknowledged/committed data was lost, corrupted, or silently wrong
- [ ] The process crashed, hung, or failed to start
- [ ] A security or isolation boundary was crossed
- [ ] Core functionality is broken with no acceptable workaround
- [x] A workaround exists (spell the types INT / INTEGER / BIGINT)

### Proposed severity

SEV-3 — Medium: DDL-surface inconsistency and Postgres-compat gap; clean workaround exists.

### Reproducibility

[...]

PR #224, opened 2026-07-26
fix(pgwire): preserve declared integer width in RowDescription OIDs; accept INT4/INT8/SMALLINT/INT2 in strict DDL

### LINKED ISSUE #217: Integer DDL types collapse wire OID fidelity — INT/INTEGER/INT4/BIGINT all report OID 20; SMALLINT/INT2 report OID 25 (text)

`farhansyah/nodedb:0.4.0` (published Docker image; `SELECT version()` → NodeDB 0.4.0, wire format 1). **Not re-tested against current `origin/main`** — close if already addressed.

### Deployment mode

Origin — single node (local), observed at the wire level through a pgwire client (Crystal / nodedb.cr, decoding `RowDescription` OIDs directly).

### Engine(s) involved

pgwire / client type fidelity — engine-independent DDL type mapping.

### Summary

Every integer DDL keyword except `SMALLINT`/`INT2` reports wire type **OID 20** (`int8`/`bigint`) in `RowDescription`, regardless of the declared width — `INT`, `INTEGER`, `INT4`, `BIGINT`, and `INT8` all come back as OID 20. `SMALLINT`/`INT2` are worse: they report **OID 25** (`text`), not `int2`, and the payload arrives as plain decimal text. A pgwire client that decodes strictly by the reported OID therefore (a) reads every `INT`/`INTEGER` column as 8-byte where it declared 4-byte, and (b) gets a text value where it expects a 2-byte integer for `SMALLINT`. The declared type and the wire-reported type disagree.

### Steps to reproduce

```sql
DROP COLLECTION IF EXISTS fixprobe_ints;
CREATE COLLECTION fixprobe_ints (id TEXT PRIMARY KEY, a INT, b INTEGER, c INT4, d BIGINT, e INT8, f SMALLINT, g INT2);
INSERT INTO fixprobe_ints (id, a, b, c, d, e, f, g) VALUES ('r1', 1, 2, 3, 4, 5, 6, 7);
SELECT id, a, b, c, d, e, f, g FROM fixprobe_ints;
```

Inspect the `RowDescription` type OIDs on the wire (psql `\gdesc`, or any client that exposes column OIDs):

| Column | DDL type | Reported OID | Reported as |
| --- | --- | --- | --- |
| `a` | `INT`      | **20** | int8/bigint |
| `b` | `INTEGER`  | **20** | int8/bigint |
| `c` | `INT4`     | **20** | int8/bigint |
| `d` | `BIGINT`   | **20** | int8/bigint |
| `e` | `INT8`     | **20** | int8/bigint |
| `f` | `SMALLINT` | **25** | text (not int2!) |
| `g` | `INT2`     | **25** | text (not int2!) |

Re-verified on a second throwaway collection (`f SMALLINT, g INT2` only, values `100, 200`) — same result: `oid=25` for both, row values arrive as plain decimal text (`"100"`, `"200"`), no quoting/JSON wrapping.

### Expected behavior

`RowDescription` OIDs should reflect the declared width: `INT`/`INTEGER`/`INT4` → OID 23 (`int4`), `BIGINT`/`INT8` → OID 20 (`int8`), `SMALLINT`/`INT2` → OID 21 (`int2`). A Postgres-wire-compatible server's whole value proposition is that a stock pg client can decode by OID; today a client must special-case NodeDB (treat all `INT` as `int8`, and parse `SMALLINT` from text).

### Actual behavior

All 4-byte and 8-byte integer keywords collapse to OID 20; `SMALLINT`/`INT2` fall through to OID 25 text. Values are still correct once decoded with this knowledge, so it is a type-fidelity/compat defect, not data loss.

### Proposed severity

SEV-3 — Medium: breaks strict-by-OID pgwire clients (they read wrong widths or get text for `SMALLINT`), but values decode correctly once the client special-cases it. Workaround: decode all integer columns as `int8`, and `SMALLINT`/`INT2` as text.

### Reproducibility

Always — every attempt.

### Environment & logs

Linux x86_64, `farhansyah/nodedb:0.4.0` Docker image, single local node. Observed by decoding `RowDescription` OIDs directly through `nodedb.cr`'s wire client. The `nodedb.cr` shard now documents "read NodeDB `INT` as `Int64`, `SMALLINT`/`INT2` as text" as a known dialect quirk in its wire-facts.

emanzx:
Fixes #217. Also fixes the DDL-acceptance half of #223.

## What

Integer columns on schemaless/columnar collections now report their declared width faithfully in pgwire `RowDescription`: `INT`/`INTEGER`/`INT4` → OID 23, `BIGINT`/`INT8` → OID 20, `SMALLINT`/`INT2` → OID 21 (previously 20/20/25-text respectively). Three commits:

1. **`fix(ddl)`** — `ColumnType::from_str` accepts `INT4`/`INT8`/`SMALLINT`/`INT2` as aliases of the existing `Int64` variant, so `document_strict`/`kv` DDL stops rejecting standard Postgres spellings (#223). No new variants, no storage change; wire-width fidelity for strict engines remains out of scope (needs a declared-type hint on `ColumnDef` — noted in #223).
2. **`fix(pgwire)`** — the core fix. The declared type string already survives into the catalog for schemaless/columnar engines; the planner's `parse_type_str` then collapsed it (`INT4` bucketed with `BIGINT`; `SMALLINT`/`INT2` unrecognized → text). Changes: `SMALLINT`/`INT2` join the integer arm; the schemaless catalog branch now forwards `raw_type` (the columnar branch already did); a new `sql_data_type_to_ddl_col_type_with_raw` refines the `Int8` wire type to `Int4`/`Int2` by the declared keyword — modeled on the already-correct `field_type_to_oid` used by catalog introspection, which is why `\d` showed the right OIDs while `SELECT` didn't. The `DdlColType::Int4`/`Int2` wire variants, OID mapping, and binary encoders all pre-existed, complete but unreachable — this change produces them for the first time.
3. **`test(pgwire)`** — binary-format regression lock: extended-query typed reads (`i16`/`i32`/`i64` via tokio_postgres) assert both the advertised OIDs and the decoded values end-to-end.

## Notes for review

- **Storage untouched**: integers remain `i64` end to end; the declared width is authoritative only for the wire type. No `ColumnType`/`SqlDataType` variants added (the wide-blast alternative was considered and rejected — int2/4/8 are freely coercible in Postgres semantics, so the distinction is wire-only).
- **One pre-existing test updated**: `extended_query_columnar_typed_scan` pinned the buggy OID (`n INT` → 20); now asserts 23 with a comment.
- **Pre-existing, newly-reachable**: the binary encoders truncate out-of-range `i64` → `i32`/`i16` silently (`shape_encode.rs`). Declared widths aren't range-enforced on write, so an overflowing value in an `INT`-declared column would wrap under binary format. Pre-existing behavior, out of scope here — flagging in case a range check is preferred as a follow-up.
- `kv`/`document_strict` columns keep reporting OID 20 for all integer widths (their `ColumnInfo` carries no `raw_type`) — covered by an explicit test so the limitation is pinned, per #223.

## Tests

- New e2e wire test (`raw_simple_query`, hand-decoded RowDescription): the issue's exact 7-keyword repro asserts `[25, 23, 23, 23, 20, 20, 21, 21]` — RED before the fix reproduced the reported `[25, 20, 20, 20, 20, 20, 25, 25]` exactly.
- New binary round-trip test (above); new strict/kv DDL-acceptance smoke tests; unit tables for the refine fn (all 7 keywords + absent/unrecognized raw + non-integer passthrough), `parse_type_str`, and the `FromStr` aliases.
- `nodedb` lib suite: 4796 passed. `nodedb-types`: 601 passed. Touched integration files: 14 passed. Clippy clean, fmt clean.

### DIFF
--- a/nodedb-sql/src/error.rs
+++ b/nodedb-sql/src/error.rs
@@ -20,6 +20,21 @@ pub enum SqlError {
     #[error("type mismatch: {detail}")]
     TypeMismatch { detail: String },
 
+    /// A write supplied an integer wider than the column's declared type.
+    ///
+    /// nodedb stores every integer as an `i64`, so this is not a storage
+    /// limit — it is the constraint that makes the column's advertised wire
+    /// type honest. A column declared `INTEGER` reports OID 23, and a pgwire
+    /// client reading it in binary format decodes exactly four bytes; letting
+    /// a wider value in would mean either truncating it on read or lying about
+    /// the type. PostgreSQL rejects the same write with the same message.
+    #[error("value {value} is out of range for column '{column}' of type {declared_type}")]
+    IntegerOutOfRange {
+        column: String,
+        value: i64,
+        declared_type: &'static str,
+    },
+
     #[error("unsupported: {detail}")]
     Unsupported { detail: String },
 
--- a/nodedb-sql/src/planner/dml.rs
+++ b/nodedb-sql/src/planner/dml.rs
@@ -6,8 +6,8 @@ use nodedb_types::DatabaseId;
 use sqlparser::ast::{self};
 
 use super::dml_helpers::{
-    build_kv_insert_plan, build_vector_primary_insert_plan, convert_value_rows,
-    resolve_insert_columns,
+    build_kv_insert_plan, build_vector_primary_insert_plan, check_declared_int_ranges,
+    convert_value_rows, resolve_insert_columns,
 };
 use crate::engine_rules::{self, InsertParams};
 use crate::error::{Result, SqlError};
@@ -135,6 +135,7 @@ pub fn plan_insert(ins: &ast::Insert, catalog: &dyn SqlCatalog) -> Result<Vec<Sq
             intent,
             Vec::new(),
             info.primary_key.as_deref(),
+            &info.columns,
         );
     }
 
@@ -148,11 +149,13 @@ pub fn plan_insert(ins: &ast::Insert, catalog: &dyn SqlCatalog) -> Result<Vec<Sq
         && let Some(ref vpc) = info.vector_primary
     {
         let rows_parsed = convert_value_rows(&columns, rows_ast)?;
+        check_declared_int_ranges(&info.columns, &rows_parsed)?;
         return build_vector_primary_insert_plan(&table_name, vpc, &columns, rows_parsed);
     }
 
     // All other engines: delegate to engine rules.
     let rows = convert_value_rows(&columns, rows_ast)?;
+    check_declared_int_ranges(&info.columns, &rows)?;
     let column_defaults: Vec<(String, String)> = info
         .columns
         .iter()
@@ -218,6 +221,7 @@ pub fn plan_upsert(ins: &ast::Insert, catalog: &dyn SqlCatalog) -> Result<Vec<Sq
             KvInsertIntent::Put,
             Vec::new(),
             info.primary_key.as_deref(),
+            &info.columns,
         );
     }
 
@@ -226,6 +230,7 @@ pub fn plan_upsert(ins: &ast::Insert, catalog: &dyn SqlCatalog) -> Result<Vec<Sq
     let columns = resolve_insert_columns(columns, &info, rows_ast)?;
 
     let rows = convert_value_rows(&columns, rows_ast)?;
+    check_declared_int_ranges(&info.columns, &rows)?;
     let column_defaults: Vec<(String, String)> = info
         .columns
         .iter()
@@ -294,6 +299,7 @@ fn plan_upsert_with_on_conflict(
             KvInsertIntent::Put,
             on_conflict_updates,
             info.primary_key.as_deref(),
+            &info.columns,
         );
     }
 
@@ -302,6 +308,7 @@ fn plan_upsert_with_on_conflict(
     let columns = resolve_insert_columns(columns, &info, rows_ast)?;
 
     let rows = convert_value_rows(&columns, rows_ast)?;
+    check_declared_int_ranges(&info.columns, &rows)?;
     let column_defaults: Vec<(String, String)> = info
         .columns
         .iter()
--- a/nodedb-sql/src/planner/dml_helpers.rs
+++ b/nodedb-sql/src/planner/dml_helpers.rs
@@ -25,6 +25,85 @@ pub(super) fn convert_value_rows(
         .collect()
 }
 
+/// Reject any integer value that does not fit its column's declared width.
+///
+/// nodedb stores every integer as an `i64`, so this is not a storage limit.
+/// It is the constraint that makes the column's advertised wire type honest:
+/// a column declared `INTEGER` reports OID 23 in `RowDescription`, and a
+/// pgwire client reading it in binary format decodes exactly four bytes.
+/// Accepting a wider value would force a later choice between truncating it on
+/// read and lying about the column's type — so the value is refused at the
+/// point it enters, exactly as PostgreSQL refuses it.
+///
+/// This runs in the planner rather than in each engine because the declared
+/// width is engine-independent (the same `IntWidth` drives the wire type for
+/// schemaless, columnar, strict, and kv alike), and because parameters are
+/// bound into the AST before planning — so one check here covers both literal
+/// `VALUES` and `$1` placeholders, for every engine, on every DML path.
+///
+/// Non-integer values and columns with no declared width pass through: this
+/// checks range only, never type.
+pub(super) fn check_declared_int_ranges(
+    columns: &[ColumnInfo],
+    rows: &[Vec<(String, SqlValue)>],
+) -> Result<()> {
+    // Overwhelmingly the common case — skip the per-cell name lookup entirely
+    // when the collection declares no narrowed integer column.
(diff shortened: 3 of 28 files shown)


farhan-syah:
Reviewed, and pushed a follow-up commit (`1d2c82b`) that resolves the review findings. The diagnosis and approach here were right — storage stays `i64`, no new `ColumnType`/`SqlDataType` variants, narrowing is wire-only. The changes below build on that rather than replacing it.

## The blocking issue: silent truncation is now reachable

The PR notes that `shape_encode.rs` truncates out-of-range `i64` → `i32`/`i16` and flags it as pre-existing and out of scope. The pre-existing/introduced distinction doesn't quite hold: those arms were **dead** for schemaless/columnar before this change, because every integer advertised `Int8`. Narrowing the OID is what promotes the declared width from decorative to load-bearing, and there is no write-side guard anywhere on that path — `SchemalessRules` discards `column_schema`, and this PR's own alias change makes `INT4` resolve to `ColumnType::Int64`, so the width never becomes a range constraint.

Concretely, before the follow-up: a schemaless column declared `INT` holding `9876543210` returned `-1294967296` under binary format, silently. Under text format it was a client-side decode failure instead.

The fix makes the declared width a real constraint, in both directions:

- **Write side** — `check_declared_int_ranges` in the planner rejects out-of-range integers with a typed `SqlError::IntegerOutOfRange` (`value 9999999999 is out of range for column 'n' of type integer`), matching what PostgreSQL raises. One choke point covers every engine and both inline literals and bound parameters, since parameters are bound into the AST before planning. KV returns early from all three INSERT/UPSERT entry points, so its check lives inside `build_kv_insert_plan` — a fourth entry point can't silently skip it.
- **Read side** — `shape_encode.rs` no longer casts with `as`. It range-checks first and raises SQLSTATE `22003` (`numeric_value_out_of_range`). Writes through SQL make this unreachable, but rows written before a width was declared, or arriving over a non-SQL ingest path, are not covered by the write check — and those must error rather than corrupt a value in flight.

## Duplicated keyword table

`sql_data_type_to_ddl_col_type_with_raw` copied `starts_with_type` and the integer arms verbatim out of `field_type_to_oid`. The doc comment said the two must agree, which is the argument for not having two of them — drift there reintroduces exactly the `\d`-says-23-but-`SELECT`-says-20 split this PR fixes.

Both now resolve through a single `nodedb_types::columnar::IntWidth`: one keyword parser, one range, one OID mapping. The refiner takes an already-resolved `IntWidth` rather than re-parsing a string, so the wire type and the write constraint cannot disagree.

(Your note that the arm ordering isn't load-bearing was correct — I verified the boundary check does stop `int` prefix-matching `int8`/`int2`. There's now a unit test pinning that directly, so a future reordering can't regress it.)

## strict/kv fidelity — fixed rather than deferred

The PR accepts `SMALLINT` on strict/kv and then reports OID 20, pinning that with a test. That trades a loud DDL error for a silent width mismatch: an ORM emitting `SMALLINT` used to get a hard failure and would now get a quiet disagreement.

This turned out not to need the `ColumnDef` change the PR anticipated, and no persisted-format change at all. `StoredCollection::fields` already records the raw declared type string for **every** engine — `create::build` fills it from the raw column list before the typed schema is built — so strict and kv resolve widths through the same path as schemaless and columnar. `ColumnInfo` now carries a resolved `int_width` instead of the `raw_type` side-channel, which also removes an unintended effect of the original patch: forwarding `raw_type` on the schemaless branch was silently populating `column_schema` for schemaless INSERT/UPSERT/UPDATE (harmless today only because `SchemalessRules` discards it).

That test's premise is inverted accordingly, and its `if let Some(col)` OID assertions are now hard lookups — as written they would pass silently if the columns stopped appearing in `RowDescription`, which is the regression they exist to catch.

## Two bugs this surfaced

- **`ALTER COLUMN ... TYPE` never updated `fields`.** Widening `INT` → `BIGINT` is the *only* change that ALTER supports (cross-type needs a rewrite and is refused), so with declared widths meaningful it was a silent no-op — the column kept advertising OID 23 and kept rejecting the values the widening was performed to allow. `fields` staleness also meant `\d` reported stale types after any ALTER, which was already wrong. Fixed across all four strict ALTER ops (type / rename / drop / add).
- **`i64::MIN` literals silently became floats.** A negative literal arrives as unary minus over a *positive* number, so `-9223372036854775808` has an operand that doesn't fit `i64`; it fell through to `Float` and lost exactness. Fixed by folding the sign into the literal before parsing.

## Smaller items

- `INT64` was still unlisted in `parse_type_str` and fell through to `_ => String` → OID 25 — the same failure mode as the `SMALLINT` bug being fixed, one line away. Added.
- `extended_query_kv_integer_value` and two tests in `pgwire_extended_query` pinned the old `INT` → OID 20 behavior, same as the columnar test this PR already updated. Updated with comments explaining the change.

## Validation

`cargo fmt` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, stage-1 suite **12338 passed / 0 failed**.

New coverage: `pgwire_int_width_range_enforcement.rs` (rejection at both ends of each width, boundary values accepted, bound parameters, UPDATE, strict + kv, and the ALTER-widens-then-accepts path), plus unit tables for `IntWidth`.

PR #225, opened 2026-07-26
feat(sql): accept SEARCH ... USING VECTOR in subquery position and with quoted collection names

### LINKED ISSUE #211: SEARCH ... USING VECTOR() cannot be used as a derived table or IN-predicate subquery

origin/main @ 7bd4d24b6

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Vector

### Summary

`SEARCH ... USING VECTOR(...)` is only accepted as a top-level statement. It cannot be used as a derived table (`SELECT * FROM (SEARCH ...) s`) or inside an `IN (...)` predicate — both fail in the SQL parser — so a k-NN result cannot be joined or filtered in the same statement. Every hybrid vector-plus-relational query therefore needs two round trips, and any relational filter has to be applied after the k-NN cut, which silently changes the result set size a caller asked for.

### Steps to reproduce

```sql
DROP COLLECTION IF EXISTS vec_probe;
CREATE COLLECTION vec_probe;
CREATE VECTOR INDEX idx_vec_probe ON vec_probe (embedding) METRIC COSINE DIM 3;
INSERT INTO vec_probe (id, title, embedding) VALUES
  ('a1','one',ARRAY[0.1,0.2,0.3]), ('a2','two',ARRAY[0.4,0.5,0.6]);

SEARCH vec_probe USING VECTOR(embedding, ARRAY[0.1,0.2,0.3], 2);
--  id | _surrogate |      distance
--  a1 | 210        | 0.0
--  a2 | 211        | 0.27000001072883606      (2 rows)   <- correct

SELECT * FROM (SEARCH vec_probe USING VECTOR(embedding, ARRAY[0.1,0.2,0.3], 2)) s;
-- ERROR:  parse error: sql parser error: Expected: joined table, found: USING at Line: 1, Column: 29

SELECT id FROM vec_probe WHERE id IN (SEARCH vec_probe USING VECTOR(embedding, ARRAY[0.1,0.2,0.3], 2));
-- ERROR:  parse error: sql parser error: Expected: ), found: vec_probe at Line: 1, Column: 42
```

### Expected behavior

A `SEARCH` result is a relation, so it composes: usable as a derived table, joinable against the base collection or another collection, and usable as the right-hand side of `IN (...)` when it projects a single column.

### Actual behavior

Both forms fail in the parser before planning. The k-NN result can only be materialized client-side and correlated with a second query, which means post-filtering happens outside the engine: `SEARCH ... k = 10` followed by a client-side filter can return far fewer than 10 usable rows with no way to ask the engine for "10 after filtering".

### Proposed severity

### Reproducibility

Always — every attempt

(blank — never worked)

### Environment & logs

Linux x86_64, release build from source (`nodedb 0.4.0`, `git commit: 7bd4d24b6`). Single local node, fresh data directory. Errors are client-visible parse errors; no server log output.

Files (best guess): `SEARCH` is dispatched as a DSL statement before the SQL parser sees a relation, so the grammar has no production placing it in a `FROM` item or subquery position.

### Before submitting

### LINKED ISSUE #212: SEARCH rejects a quoted collection name while accepting a quoted column argument

origin/main @ 7bd4d24b6

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Vector

### Summary

`SEARCH "collection" USING VECTOR(...)` fails in the parser — the collection name in a `SEARCH` statement must be a bare identifier. A double-quoted column argument inside `VECTOR(...)` is accepted, so quoting is handled inconsistently within the same statement. A collection whose name requires quoting (mixed case, or a name colliding with a keyword) therefore cannot be searched at all.

### Steps to reproduce

```sql
DROP COLLECTION IF EXISTS vec_probe;
CREATE COLLECTION vec_probe;
CREATE VECTOR INDEX idx_vec_probe ON vec_probe (embedding) METRIC COSINE DIM 3;
INSERT INTO vec_probe (id, title, embedding) VALUES
  ('a1','one',ARRAY[0.1,0.2,0.3]), ('a2','two',ARRAY[0.4,0.5,0.6]);

SEARCH vec_probe USING VECTOR(embedding, ARRAY[0.1,0.2,0.3], 2);
-- 2 rows      <- bare names: correct

SEARCH vec_probe USING VECTOR("embedding", ARRAY[0.1,0.2,0.3], 2);
-- 2 rows      <- quoted column: accepted

SEARCH "vec_probe" USING VECTOR(embedding, ARRAY[0.1,0.2,0.3], 2);
-- ERROR:  parse error: sql parser error: Expected: an SQL statement, found: SEARCH at Line: 1, Column: 1
```

### Expected behavior

A double-quoted identifier is accepted wherever a bare identifier is, including the collection position of `SEARCH`. Consistent with the quoted column argument already being accepted.

### Actual behavior

The statement is rejected at column 1 — the dispatcher does not recognize the statement as a `SEARCH` at all once the collection name is quoted, so the error text points at `SEARCH` rather than at the quoting. Collections whose names need quoting are unreachable through vector search.

### Proposed severity

### Reproducibility

Always — every attempt

(blank — never worked)

### Environment & logs

Linux x86_64, release build from source (`nodedb 0.4.0`, `git commit: 7bd4d24b6`). Single local node, fresh data directory. Client-visible parse error only.

Files (best guess): the DSL statement dispatcher matches on the raw token sequence before identifier unquoting, so a quoted collection name breaks the prefix match.

### Before submitting

mkhairi:
Closes #211
Closes #212

## What this changes

Two parse-surface defects in `SEARCH ... USING VECTOR`, plus two silent
wrong-result bugs underneath them that the first fix made reachable.

### #211 — `SEARCH` could not be used as a subquery

The clause was only accepted as a whole statement, so a k-NN result could not be
filtered or joined in the same query:

```sql
SELECT * FROM (SEARCH docs USING VECTOR(embedding, ARRAY[0.1,0.2,0.3], 2)) s;
-- ERROR: parse error: sql parser error: Expected: joined table, found: USING
```

`SEARCH` is now also recognised in subquery position, so the result composes as
a relation:

```sql
SELECT id FROM (SEARCH docs USING VECTOR(embedding, ARRAY[0.1,0.2,0.3], 2)) s
WHERE s.tag = 'keep';
```

### #212 — a quoted collection name was rejected

`SEARCH "docs" USING VECTOR(...)` failed at column 1 because the collection
position accepted only bare identifier characters, while a quoted *column*
argument was already accepted. Quoted names are now taken in both positions,
with the doubled-quote escape (`""`) honoured, and the quotes are preserved
through the rewrite so a name that needs them still resolves.

### Two silent wrong-result bugs found underneath

Both are reachable today with plain SQL and no `SEARCH` involved. They are
fixed here because the new syntax makes them trivial to hit.

1. **An outer `LIMIT` over any CTE or derived table was dropped.** `apply_limit`
   had no `SqlPlan::Cte` arm, so the clause fell into `_ => {}`:

   ```sql
   WITH k AS (SELECT * FROM docs LIMIT 100) SELECT id FROM k LIMIT 1;  -- 100 rows
   ```

2. **Outer filters and projection over a non-`Scan` CTE body were dropped.**
   `inline_cte` merged outer constraints only when the body was a `Scan`; any
   other body returned `cte_plan.clone()` and the outer `WHERE` vanished:

   ```sql
   WITH k AS (SELECT * FROM docs ORDER BY vector_distance(emb, ARRAY[0.1,0.2]) LIMIT 2)
   SELECT id FROM k WHERE id = 'a1';   -- returned both rows
   ```

   Filters, projection, and an unordered `LIMIT` now merge onto a
   `VectorSearch` body. Other non-`Scan` bodies (`Aggregate`, `Join`,
   `TextSearch`, ...) still drop outer constraints — untouched here, see
   follow-ups.

## Semantics worth confirming

With the predicate merged into the search, `k` counts *matching* rows:

```sql
-- r0, r1 are the two nearest; r0 and r2 are tagged 'keep'
SELECT id FROM (SEARCH docs USING VECTOR(emb, q, 2)) s WHERE s.tag = 'keep';
-- this PR:    r0, r2   (the 2 nearest rows that match)
-- strict SQL: r0       (the 2 nearest, then filtered)
```

The engine-side behaviour is the useful one — it is the only way to ask for "k
results after filtering", which is what #211 is about — but it does deviate
from derived-table semantics, and `SqlPlan::VectorSearch` has no slot for
"top-k, then filter". Flagging it rather than burying it: if strict SQL
semantics are preferred, that shape needs a post-filter node above the search,
which is a larger change. Happy to follow whichever you prefer.

## Implementation

- `nodedb-sql/src/parser/preprocess/search_vector.rs` — the rewrite is factored
  into `parse_search_clause`, returning the canonical `SELECT` plus the bytes it
  consumed. `try_rewrite_nested_search_using_vector` walks the literal-aware
  paren positions from `lex::find_operator_positions` and rewrites every
  `(SEARCH ...)` it finds. `take_identifier` handles quoted names. Keyword
  matching is byte-wise so a multi-byte literal after `(` cannot slice
  mid-character.
- `nodedb-sql/src/parser/preprocess/pipeline.rs` — the nested pass runs after
  the top-level one, gated on the word `SEARCH` appearing at all.
- `nodedb-sql/src/planner/select/entry.rs` — `apply_limit` descends into
  `SqlPlan::Cte`'s outer plan.
- `nodedb/src/control/planner/sql_plan_convert/expr.rs` — `inline_cte` merges
  outer filters/projection/`LIMIT` onto a `VectorSearch` body.

The top-level path got slightly stricter as a side effect: the old
`strip_parentheses` only checked the first and last character, so
`SEARCH t USING VECTOR(emb, ARRAY[1], 3) AND (y)` was accepted and mangled. It
is now rejected (`top_level_rewrite_rejects_trailing_clause`).

## How to test

```bash
cargo nextest run -p nodedb --test sql_search_subquery_composition   # 6 tests
cargo nextest run -p nodedb-sql --lib parser::preprocess             # 101 tests
cargo nextest run -p nodedb --lib control::planner::sql_plan_convert::expr
```

Integration coverage: derived table returns the k nearest in distance order; an
outer `WHERE` reaches the engine; an outer `LIMIT` narrows the cut; an outer
projection narrows the columns; a quoted collection name searches the same
collection as the bare form, bare and in subquery position; a malformed
argument list is rejected rather than rewritten.

Unit coverage: nested rewrite for one and several occurrences, string-literal
and comment safety, multi-byte input, quoted and escaped-quote names,
unterminated and empty quoted names, `FUSION` left alone, trailing-clause
rejection; `inline_cte` filter merge, `LIMIT` narrowing, and `LIMIT` above `k`
leaving the cut untouched.

## Tradeoffs and alternatives

- **A textual rewrite rather than a grammar production.** `SEARCH ... USING
  VECTOR` is already a preprocess rewrite, so extending it keeps one
  implementation. A real `TableFactor` production would be the cleaner
  long-term home; this change does not block it.
- **`SEARCH ... USING FUSION(...)` and the other DSL forms are untouched** —
  they are not rewritten to SQL, so they remain statement-only.
- **Bare `IN (SEARCH ...)` is still a parse error.** `SEARCH` projects `id`,
  `_surrogate`, and `distance`, so it can never be a valid `IN` operand; the
  composable form is `IN (SELECT id FROM (SEARCH ...) s)`.

## Follow-ups (not in this PR)

- A semi-join whose inner side is a vector search returns no rows:
  `SELECT id FROM docs WHERE id IN (SELECT id FROM docs ORDER BY vector_distance(emb, q) LIMIT 2)`
  yields 0 rows. Reproducible without `SEARCH`; needs work in the join
  execution path.
- Outer constraints over non-`Scan`, non-`VectorSearch` CTE bodies are still
  dropped, and an outer `ORDER BY`/`OFFSET`/`DISTINCT` over a `VectorSearch`
  body has no slot to land in.

### DIFF
--- a/nodedb-cluster-tests/tests/transport_security.rs
+++ b/nodedb-cluster-tests/tests/transport_security.rs
@@ -38,12 +38,12 @@ fn insecure_counter_guard() -> &'static Mutex<()> {
     LOCK.get_or_init(|| Mutex::new(()))
 }
 
+use nodedb_cluster::topology::{ClusterTopology, NodeInfo, NodeState};
 use nodedb_cluster::transport::pki_types::{CertificateDer, PrivateKeyDer, PrivatePkcs8KeyDer};
 use nodedb_cluster::{
     MacKey, NexarTransport, RaftRpcHandler, TlsCredentials, TransportCredentials,
     generate_node_credentials, insecure_transport_count, spki_pin_from_cert_der,
 };
-use nodedb_cluster::topology::{ClusterTopology, NodeInfo, NodeState};
 use nodedb_raft::message::{AppendEntriesRequest, AppendEntriesResponse, RequestVoteResponse};
 use nodedb_raft::transport::RaftTransport;
 
--- a/nodedb-physical/src/physical_plan/mod.rs
+++ b/nodedb-physical/src/physical_plan/mod.rs
@@ -324,6 +324,8 @@ impl PhysicalPlan {
             | PhysicalPlan::Graph(GraphOp::WccSuperstep(_)) => None,
             // Exchange: recurse into the child plan to extract the collection.
             PhysicalPlan::Query(QueryOp::Exchange(op)) => op.child.collection(),
+            // PostProcess: recurse into the materialized input plan.
+            PhysicalPlan::Query(QueryOp::PostProcess { input, .. }) => input.collection(),
             // ProviderScan is a catalog/constant source — no user collection.
             PhysicalPlan::Query(QueryOp::ProviderScan { .. }) => None,
             // KV ops carry their own collection (sorted-index-only ops → None).
--- a/nodedb-physical/src/physical_plan/query.rs
+++ b/nodedb-physical/src/physical_plan/query.rs
@@ -92,6 +92,46 @@ pub enum QueryOp {
         distinct: bool,
     },
 
+    /// Relational post-processing over a materialized child plan.
+    ///
+    /// Coordinator-resolved, exactly like [`QueryOp::Exchange`]: at resolve time
+    /// the coordinator gathers `input`'s rows, flattens them to the relational
+    /// row shape, and rewrites this node into a [`QueryOp::ProviderScan`]
+    /// carrying the same relational parameters — which applies
+    /// filter→offset→sort→distinct→project→limit on a single core. A Data-Plane
+    /// core must NEVER see this node (defensive error if it does).
+    ///
+    /// Lowered from `SqlPlan::Subquery`: an outer `ORDER BY` / `OFFSET` /
+    /// `DISTINCT` / post-reorder `LIMIT` over a subquery/derived-table body
+    /// whose leaf plan could not absorb those constraints. Constraints the leaf
+    /// DID absorb (a `WHERE` pushed into a search engine, an unordered `LIMIT`
+    /// folded into `top_k`) are applied by `input` and not repeated here.
+    PostProcess {
+        /// The subquery body to materialize. Wrapped in `Exchange{Gather}` by
+        /// the converter when the body is a sharded source, so the gather runs
+        /// exactly once over the full union before post-processing.
+        input: Box<crate::physical_plan::PhysicalPlan>,
+        /// Serialized `Vec<ScanFilter>` (MessagePack). Empty = no predicate.
+        #[serde(default)]
+        filters: Vec<u8>,
+        /// Output column names to keep. Empty = emit all columns.
+        #[serde(default)]
+        projection: Vec<String>,
+        /// Sort keys: `(column_name, ascending)`. Empty = unordered.
+        #[serde(default)]
+        sort_keys: Vec<(String, bool)>,
+        /// Maximum rows to emit after offset/sort/distinct/projection.
+        /// `None` = unlimited.
+        #[serde(default)]
+        limit: Option<usize>,
+        /// Number of rows to skip before applying limit.
+        #[serde(default)]
+        offset: usize,
+        /// SQL DISTINCT: deduplicate on the projected row.
+        #[serde(default)]
+        distinct: bool,
+    },
+
     /// Aggregate: GROUP BY + aggregate functions.
     Aggregate {
         collection: String,
--- a/nodedb-physical/src/physical_plan/routing.rs
+++ b/nodedb-physical/src/physical_plan/routing.rs
@@ -75,6 +75,13 @@ pub fn plan_contains_cluster_partitioned_leaf(plan: &PhysicalPlan) -> bool {
             plan_contains_cluster_partitioned_leaf(child)
         }
 
+        // Recurse through PostProcess (its materialized input is the real
+        // plan). Normally resolved to a `ProviderScan` before routing, but a
+        // conservative recursion keeps routing correct if one survives.
+        PhysicalPlan::Query(QueryOp::PostProcess { input, .. }) => {
+            plan_contains_cluster_partitioned_leaf(input)
+        }
+
         // Recurse through lateral outer plans.
         PhysicalPlan::Query(QueryOp::LateralTopK { outer_plan, .. })
         | PhysicalPlan::Query(QueryOp::LateralLoop { outer_plan, .. }) => {
--- a/nodedb-sql/src/parser/preprocess/pipeline.rs
+++ b/nodedb-sql/src/parser/preprocess/pipeline.rs
@@ -13,7 +13,9 @@
 use super::function_args::rewrite_object_literal_args;
 use super::lex::{first_sql_word, has_brace_outside_literals, has_operator_outside_literals};
 use super::object_literal_stmt::try_rewrite_object_literal;
-use super::search_vector::try_rewrite_search_using_vector;
+use super::search_vector::{
+    try_rewrite_nested_search_using_vector, try_rewrite_search_using_vector,
+};
 use super::temporal::extract as extract_temporal;
 use super::vector_ops::{
     rewrite_arrow_distance, rewrite_cosine_distance, rewrite_neg_inner_product,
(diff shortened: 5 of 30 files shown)


farhan-syah:
Rebased onto the latest `main` and added a follow-up commit that turns two silent wrong-result bugs this PR made reachable into correct behaviour.

## What the follow-up adds

An outer `ORDER BY` / `OFFSET` / `DISTINCT` (and a `LIMIT` that must apply *after* a reorder) over a subquery/derived-table body whose leaf plan can't absorb it was previously dropped — returning the body's unordered/unoffset/undeduplicated rows. This adds a generic relational **post-processing operator** so those constraints execute:

- `SqlPlan::Subquery` → `QueryOp::PostProcess`, a coordinator-resolved node (like `Exchange`): it gathers the child once across shards, flattens to relational rows, and lowers to the existing `ProviderScan` tail (filter → offset → sort → distinct → project → limit). Sort runs before projection, so `ORDER BY` may reference a non-selected column.
- The vector-search fast path is preserved: an outer `WHERE` still merges into the engine post-filter (so `k` counts matching rows), and an unordered `LIMIT` still folds into `top_k`. The wrapper only handles what the leaf genuinely can't.
- **Vector payload materialization**: a hit's document `body` is flattened to top-level columns (with `distance` / `_surrogate`, and the document's own `id` winning), so ordering/distinct/projection over a `SEARCH` subquery works for any column, not just `id` / `distance`.
- Security-sensitive recursion into the new node: RLS injection, authorization requirements, and required-permission all descend into the post-processor's child, so a subquery body cannot bypass RLS or authz.

## Coverage

New integration tests: outer `ORDER BY` on a payload column reorders the k-NN result; `OFFSET` skips leading rows; `DISTINCT` dedups the projected column; `ORDER BY distance DESC LIMIT` takes the farthest. Plus unit tests for the hit flatten and the `inline_cte` wrap/no-wrap decision.

## Validation

`cargo fmt` clean · `cargo clippy --all-targets -- -D warnings` clean · stage-1 suite (`--workspace --exclude nodedb-cluster-tests --all-features`) **12,561 passed, 0 failed** · cluster suite **316 passed, 0 failed**.

## Follow-up (not in this PR)

Text / hybrid / spatial subquery bodies order/distinct correctly by `score` / `id`, but ordering by an *unmaterialized payload column* over those engines still leaves the document body nested — extending the hit flatten to them is the natural next step.

PR #226, opened 2026-07-26
fix(sql): raise 42883/22012 for undefined functions and division by zero instead of folding to NULL

### LINKED ISSUE #216: Unknown function call and `1/0` return one NULL row instead of erroring (42883 / 22012 suppressed)

`farhansyah/nodedb:0.4.0` (published Docker image; `SELECT version()` → NodeDB 0.4.0, wire format 1). **Not re-tested against current `origin/main`** — close if already addressed.

### Deployment mode

Origin — single node (local), observed through a pgwire client (Crystal / nodedb.cr).

### Engine(s) involved

SQL semantics (parser/evaluator) — engine-independent.

### Summary

A call to an **unknown function** evaluates as a *successful* query returning one row with a single `NULL` column, instead of raising `42883 undefined_function` the way Postgres does. `SELECT 1/0` behaves the same way — a successful one-row `NULL`, not a `22012 division_by_zero`. Because the failure is reported as success with a NULL payload, a client cannot distinguish "this expression errored" from "this expression legitimately evaluated to NULL". This suppresses a whole class of errors silently — a typo'd function name, a bad cast, or a divide-by-zero all look like a valid NULL result.

### Steps to reproduce

```sql
-- unknown function: Postgres raises 42883; NodeDB returns one NULL row, no error
SELECT some_function_that_does_not_exist(1, 2);
--  (1 row)
--  NULL

-- division by zero: Postgres raises 22012; NodeDB returns one NULL row, no error
SELECT 1/0;
--  (1 row)
--  NULL
```

For contrast, a genuinely malformed reference *does* error correctly — e.g. `SELECT * FROM nonexistent_table_xyz` returns `ERROR 42P01 collection "nonexistent_table_xyz" does not exist` — so the error path exists; it is specifically unknown-function and arithmetic-domain evaluation that is swallowed.

### Expected behavior

An unknown function should raise `42883` (undefined_function); `1/0` should raise `22012` (division_by_zero). At minimum, an expression that could not be evaluated must not be reported to the client as a successful `NULL` result — the client has no way to tell the two apart, which is exactly the condition that lets a bug (or an injected/hallucinated query) pass silently.

### Actual behavior

Both cases return `CommandComplete` with one row and a single `NULL` (text/OID 25) column, no `ErrorResponse`. The evaluator appears to fold an unresolved/failed scalar expression to `NULL` rather than surfacing an error.

### Proposed severity

SEV-3 — Medium (flagging a possible security angle for triage): silently-wrong results rather than data loss, but the error-suppression class is a real footgun — it hides typos, bad casts, and injected/malformed expressions from any client that relies on an error to detect them. Maintainers may weigh whether the suppression warrants higher at triage, given the security-hardening context.

### Reproducibility

Always — every attempt.

### Environment & logs

Linux x86_64, `farhansyah/nodedb:0.4.0` Docker image, single local node. Observed via a Crystal pgwire client while establishing a wire-facts baseline for `nodedb.cr` — this behavior is why the ORM asserts *result rows / server state* in every dialect probe rather than trusting absence-of-error (a `SELECT`'d expression "not erroring" is not evidence it worked).

emanzx:
Fixes #216.

## What

Failed scalar expressions stop masquerading as successful NULL results:

- **Unknown function → SQLSTATE 42883** at plan time. Root cause: nothing validated function existence anywhere — the runtime evaluator's dispatch chain ends in `unwrap_or(Value::Null)`. A registry gate now rejects unknown names in the resolver (after the FTS/catalog interceptions), mirroring the existing undefined-table (42P01) timing.
- **Division/modulo by zero → SQLSTATE 22012** at runtime, across integer, float, and Decimal paths, both the `%`/`/` operators and the `mod()` function. Root cause: the evaluator was infallible *by type* (`eval() -> Value`) — the zero-divisor arms hand-returned NULL. Evaluation is now `Result<Value, EvalError>`, propagated through every reachable call site (scan filters, projections, write paths, enforcement, overlay merges, cold-tier predicate pushdown, streaming and non-streaming pgwire).

## Registry completeness (part of the fix, not scope creep)

Gating unknown names exposed that the SQL-side function registry only knew a fraction of the runtime evaluator's implemented functions — **80 names registered** across math, string, conditional, id, array, json-legacy, datetime, and types (plus a 27-name spatial audit), each with arg metadata derived from the implementation. A new invariant test transcribes every runtime dispatch module's match arms (including geo) and asserts registry coverage, so the two sides can no longer drift silently.

## Deliberate behavior changes (Postgres-correct, called out for review)

- **WHERE clauses**: a row whose predicate divides by zero used to be silently filtered (NULL ≈ false); the query now errors with 22012. Pinned by tests on both document and columnar engines.
- **Extended protocol**: `Parse` succeeds for a statement calling an unknown function — planning is deferred, so 42883 surfaces at Bind/Execute rather than Parse (unlike undefined-table, which rejects at Parse). Documented in the new extended-protocol tests; flagging in case Parse-time planning is preferred long-term.

## Deliberately still folding to NULL (documented in-code, each with a rationale comment)

Expressions *inside* aggregate accumulators, GROUP BY key building, window functions, the hash-join probe hot path, and the (currently unwired) trigger WHEN batch filter. These are pervasive infallible hot paths whose conversion is a materially larger change — tracked in #228. Also out of scope: CAST failures, known-function arity mismatches (registry arg metadata exists but remains unenforced), the procedural/PL evaluator (tracked in #227 — its `NEW.x := a/b` assignments still write NULL on zero divisors), and the from-less-`SELECT` constant-fold path (pre-existing separate gap; the new tests route around it via `FROM` + column divisors and document why).

## Other notes for review

- **CHECK/transition/typeguard enforcement**: an unevaluable constraint fails the statement (fails closed); on the CRDT merge path it surfaces as that check's own `Violation` shape rather than a bare 22012 — meaning a collection configured `ON CONFLICT LastWriterWins FOR CHECK` resolves an *unevaluable* predicate the same way as a genuine conflict. Kept as-is (widening `ValidationOutcome` is a cross-crate API change); worth a maintainer look.
- **Rolling upgrades**: no wire bytes changed, so old/new nodes coexist during a deploy and evaluate the same shipped expression with different semantics (old folds, new errors) until the upgrade completes.
- **Cluster fidelity**: both codes round-trip cross-node via new numeric ErrorCodes + msgpack variants, so remote errors keep their SQLSTATE instead of degrading to XX000. The streaming pgwire path needed (and got) an explicit bridge so mid-stream errors keep 22012.
- RLS predicates fail closed on eval errors (now with a warn log); the RLS grammar cannot currently construct arithmetic, so this is defensive only.

## Tests

Adversarially reviewed (three independent passes) before submission. New coverage: plan-time 42883 with zero rows scanned; 22012 across int/float/Decimal/mod, WHERE-flip on two engines, computed columns, cold-tier pushdown; CASE/COALESCE laziness pinned at unit + integration level (`CASE WHEN false THEN 1/0` still succeeds); extended-protocol coverage for both codes; registry invariant test; controls for valid functions, valid division, NULL arithmetic, and 42P01. Full targeted totals on top of current main: nodedb-sql 754, nodedb-query 459, nodedb lib 4810, new integration files 30 — zero failures; clippy clean.


farhan-syah:
Follow-up pushed (rebased on current `main`): completed the expression-error propagation that earlier revisions had scoped out, plus consistency fixes surfaced in review.

**Division/modulo-by-zero now raises SQLSTATE 22012 in every evaluated path** (previously folded to NULL in these):
- Aggregate arguments (`SUM(1/x)` etc.) — threaded through the streaming accumulator, spill, and shuffle paths
- Computed `GROUP BY` keys
- Window `ORDER BY` / `PARTITION BY` (both the JSON and value window evaluators)
- Hash-join residual `ON` predicates (including the grace-hash spill/streaming path)
- Trigger `WHEN` filter (BEFORE fails the statement; AFTER logs and skips, since it runs post-commit)

**Enforcement consistency:** transition-check and type-guard evaluation errors now map to `22012`, matching generated-column / materialized-sum enforcement, instead of each constraint's own violation code.

**CRDT validation:** added a first-class `ValidationOutcome::EvalError` so an unevaluable CHECK predicate on the sync-apply path fails closed (escalated to the DLQ) rather than being "resolved" by a conflict policy like LastWriterWins. Threaded through all consumers.

**Response-boundary fixes:** three sites (`over_docs`, `owning_core`, `gather`) were degrading the typed `DivisionByZero` code to a generic internal error (XX000) before it reached pgwire; added a central reverse-mapping helper so the code round-trips to `22012`.

Also removed stale issue-number references from code comments now that the fix spans many files.

**Validation:** rebased on `main`; `cargo fmt` + `clippy -D warnings` clean; full affected-crate suite green (9418 tests, incl. new integration coverage for the aggregate/group-by/window/join paths and unit coverage for the CRDT and enforcement changes); cluster stage-2 green (the one placement-convergence failure is a known timing flake — passes on retry, unrelated subsystem).

PR #229, opened 2026-07-27
fix(sql): reject procedural division/modulo by zero, reusing #216 error plumbing

### LINKED ISSUE #227: Trigger-body assignments silently write NULL on division by zero — procedural evaluator folds expression errors (22012 suppressed on the write path)

`origin/main @ f7fcc7718` (release build; predates #226, but the affected evaluator is untouched by that PR — see Summary).

### Deployment mode

Origin — single node (local), via psql over pgwire.

### Engine(s) involved

Procedural executor (trigger bodies / DO blocks) — a separate evaluator from the scalar-expression path fixed in #226.

### Summary

The procedural executor has its own arithmetic evaluator, and it silently folds division by zero to NULL on the write path: a `BEFORE INSERT` trigger body assignment `NEW.ratio := NEW.numerator / NEW.denominator;` with a zero denominator writes `NULL` into the row and lets the INSERT commit — no error, no SQLSTATE 22012. This is the same suppressed-error class as #216, in a different evaluator: `try_eval_constant`'s division arm only matches a non-zero divisor, the zero case falls through to `None`, and `evaluate_to_value` coerces `None` to `Ok(Value::Null)` (`control/planner/procedural/executor/eval.rs` ~181/195, `~88-92`), reaching `execute_assign` / `execute_return`. #226 deliberately scoped this out (separate evaluator); filing so it is tracked rather than implicit.

### Steps to reproduce

```sql
CREATE COLLECTION trig_probe (id TEXT PRIMARY KEY, numerator INT, denominator INT, ratio INT);
CREATE TRIGGER trig_ratio BEFORE INSERT ON trig_probe FOR EACH ROW
  BEGIN NEW.ratio := NEW.numerator / NEW.denominator; END;
INSERT INTO trig_probe (id, numerator, denominator) VALUES ('r1', 10, 0);  -- succeeds
INSERT INTO trig_probe (id, numerator, denominator) VALUES ('r2', 10, 2);  -- succeeds
SELECT id, numerator, denominator, ratio FROM trig_probe ORDER BY id;
--  r1 | 10 | 0 | NULL   <- silently wrong: 10/0 folded to NULL, write committed
--  r2 | 10 | 2 | 5      <- control: trigger works
```

### Expected behavior

The zero-divisor assignment raises division_by_zero (SQLSTATE 22012), aborting the triggering statement — consistent with the post-#226 behavior of the main scalar evaluator, and with the procedural engine's own `EXCEPTION WHEN DIVISION_BY_ZERO` handler machinery (which string-matches a phrase that is currently never raised anywhere).

### Actual behavior

The assignment silently produces NULL; the INSERT commits with the wrong value; no error is observable anywhere.

- [ ] The process crashed, hung, or failed to start
- [x] A workaround exists (guard the divisor in the trigger body with an IF, or validate before insert)

### Proposed severity

SEV-2 — High: silently-wrong stored results on the write path; pre-existing data intact.

### Reproducibility

Always — every attempt.

Not a regression — the procedural evaluator has folded eval failures to NULL since introduction.

### Environment & logs

Linux x86_64, source build of `origin/main @ f7fcc7718`, single local node, fresh collections. Found during the adversarial review pass for #226. Note: the same `None → Ok(Value::Null)` coercion in `evaluate_to_value` also swallows other expression failures in trigger bodies (unknown identifiers, unsupported operators), so a fix probably wants a typed error out of `try_eval_constant` rather than a div-zero special case.

### Before submitting

- [x] I reproduced this on a released tag or a current `main` build.

emanzx:
Fixes #227.

## What

Trigger-body assignments stop silently writing NULL on division by zero: `NEW.ratio := NEW.numerator / NEW.denominator;` with a zero denominator now aborts the triggering statement with a division-by-zero error instead of committing a wrong row. Same class as #216, but in a **different evaluator** — the procedural executor's own constant folder, which every trigger body / `CALL` / scheduled job / alert body runs through. #216 fixed the `nodedb_query` row-expression path; this fixes the procedural path, which on `main` today still folds a zero divisor to `NULL`.

Root cause was one coercion: `evaluate_to_value` mapped every failed constant-fold (`None`) to `Ok(Value::Null)` — including the zero-divisor arms that deliberately declined to divide. The fix threads a typed `ConstEvalError::DivisionByZero` through `try_eval_constant` / `eval_expr` / `eval_binary_op` so the zero-divisor cases become `crate::Error::DivisionByZero`, while every other fold-to-NULL shape (unknown identifiers, unsupported operators, overflow, non-finite float divisors) keeps today's behavior — scope-disciplined to `/` and `%` by zero only.

**Bonus that fell out for free:** the `EXCEPTION WHEN DIVISION_BY_ZERO` handler machinery — present and unit-tested since introduction, but matching an error nothing ever raised — now works end-to-end. A trigger body with `EXCEPTION WHEN DIVISION_BY_ZERO THEN NEW.ratio := 0; END` catches a real division by zero and rescues the insert (covered by a new e2e test).

## Scope / notes for review

- **Reuses #216's error plumbing.** This PR was originally opened against a base predating #216 and carried its own copy of the `nodedb-types` plumbing (SQLSTATE `22012`, `ErrorCode` 1204, msgpack tag 68, the `Error::DivisionByZero` variant). #216 has since merged that same plumbing to `main`, so the PR has been **rebased onto current `main` and the redundant 11-file plumbing dropped** — it now reuses the existing `Error::DivisionByZero` variant. The diff is the fix and its tests only.
- **Known pre-existing gap, not addressed here:** the trigger-firing wrapper flattens every inner trigger error (including `RAISE EXCEPTION`) to `BadRequest` / SQLSTATE 42601, so an *uncaught* division by zero in a trigger surfaces to the client as 42601 with "division by zero" in the message rather than 22012. Fixing that wrapper's error fidelity is a separate, broader change.
- Behavior change: any trigger/procedure relying on silent NULL-on-zero-divisor now hard-fails. No existing test pinned the old behavior.
- All live procedural consumers (BEFORE/AFTER/INSTEAD-OF triggers, `CALL`, scheduled jobs, alert bodies) share the fixed choke point; IF/WHILE conditions already hard-failed and now carry the precise error.

## Tests

New unit module for the procedural evaluator (int/float/modulo zero-divisor, non-zero control, unrelated-`None`-still-folds control) and three e2e trigger tests: the issue's exact repro (errors + row not persisted), the happy path (ratio computed and persisted), and the exception-handler rescue.

- nodedb-types error round-trip: 91 passed
- procedural eval unit tests: 5 passed
- trigger_e2e: 21 passed
- fmt clean; touched files clippy-clean

### DIFF
--- a/nodedb/src/control/planner/procedural/executor/eval.rs
+++ b/nodedb/src/control/planner/procedural/executor/eval.rs
@@ -12,6 +12,31 @@
 use crate::control::state::SharedState;
 use crate::types::TenantId;
 
+/// Signal from constant-expression evaluation that a `/` or `%` reached a
+/// zero divisor (nodedb issue #227).
+///
+/// Kept local to this module — the procedural executor's constant folder is
+/// a separate, sqlparser-based tree-walk evaluator with no dependency on
+/// `nodedb_query` (whose own `EvalError::DivisionByZero`, from the
+/// DataFusion-backed row-expression evaluator fixed under issue #216,
+/// covers a different code path entirely). Every other historically
+/// `None`-folding case (unknown identifiers, unsupported operators,
+/// overflow, non-finite float results) is unaffected and keeps folding to
+/// `Ok(None)` — see `try_eval_constant`, `evaluate_to_value`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+enum ConstEvalError {
+    #[error("division by zero")]
+    DivisionByZero,
+}
+
+impl From<ConstEvalError> for crate::Error {
+    fn from(e: ConstEvalError) -> Self {
+        match e {
+            ConstEvalError::DivisionByZero => Self::DivisionByZero,
+        }
+    }
+}
+
 /// Evaluate a SQL boolean condition.
 ///
 /// Fast path for constant TRUE/FALSE/1/0/NULL. Falls back to constant
@@ -25,7 +50,7 @@ pub async fn evaluate_condition(
         return Ok(result);
     }
 
-    match try_eval_constant(condition_sql) {
+    match try_eval_constant(condition_sql)? {
         Some(nodedb_types::Value::Bool(b)) => Ok(b),
         Some(nodedb_types::Value::Integer(i)) => Ok(i != 0),
         Some(nodedb_types::Value::Float(f)) => Ok(f != 0.0),
@@ -48,7 +73,7 @@ pub async fn evaluate_int(
         return Ok(v);
     }
 
-    match try_eval_constant(sql) {
+    match try_eval_constant(sql)? {
         Some(nodedb_types::Value::Integer(i)) => Ok(i),
         Some(nodedb_types::Value::Float(f)) => Ok(f as i64),
         _ => Err(crate::Error::PlanError {
@@ -87,7 +112,7 @@ pub async fn evaluate_to_value(
         return Ok(nodedb_types::Value::String(inner.replace("''", "'")));
     }
 
-    match try_eval_constant(trimmed) {
+    match try_eval_constant(trimmed)? {
         Some(val) => Ok(val),
         None => Ok(nodedb_types::Value::Null),
     }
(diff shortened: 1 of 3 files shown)


farhan-syah:
Rebased onto current `main` and force-pushed (`62f0d0be9`). The diff shrank from 13 files to 3 — here's why, so the reduced surface isn't mistaken for lost work.

**What changed:** this PR was opened against a base predating #216. In the meantime #216 merged the exact same `DivisionByZero` error plumbing to `main` (SQLSTATE `22012`, `ErrorCode` 1204, msgpack tag 68, the `NodeDbError::division_by_zero()` ctor, and the `Error::DivisionByZero` variant) — byte-for-byte the same, and `main` is a strict superset (it also carries tag 67 `UndefinedFunction`, which this branch had left gapped). All 11 of those plumbing files conflicted on rebase and were resolved to `main`'s version, so the PR now **reuses** the existing variant instead of re-declaring it. The one substantive merge was the `Error::DivisionByZero` doc comment, which now credits both sources sharing the variant (the `nodedb_query` row-expression evaluator from #216 **and** this procedural constant folder).

**What did *not* change:** the actual fix. #216 fixed the `nodedb_query` row-expression path; it did **not** touch the procedural executor's constant folder in `procedural/executor/eval.rs`, which on `main` today still folds a zero divisor to `NULL` and silently commits a wrong row from any trigger body / `CALL` / scheduled job. That path is exactly what this PR fixes — it's a live bug, not a duplicate of #216.

**Final diff (vs `origin/main`):**
- `procedural/executor/eval.rs` — the fix + 5 unit tests
- `tests/trigger_e2e.rs` — 3 e2e tests (repro / happy-path / exception rescue)
- `error.rs` — one-line doc-comment merge

**Validation on the rebased base:** fmt clean; touched files clippy-clean; nodedb-types error round-trip 91/91; procedural eval unit tests 5/5; trigger_e2e 21/21.

PR #240, opened 2026-08-09
fix(client): reconstruct typed errors from the server's SQLSTATE

### LINKED ISSUE #239: Native client discards the server's SQLSTATE: every server error becomes ErrorCode(9000) Internal, and retriability is lost

origin/main @ 28287607a

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Not engine-specific / unsure

### Summary

`nodedb-client`'s native protocol connection reads only the `message` field of a server error frame and drops the `code` field beside it, so every server-returned error reaches the caller as `ErrorCode(9000)` / `ErrorDetails::Internal` regardless of what the server actually classified it as. The SQLSTATE is computed by the server, is carried on the wire, and is documented on the frame type; it is discarded on arrival.

The consequence is not only a less useful error string. Every predicate on `NodeDbError` derives from `ErrorDetails`, so folding everything into `Internal` also makes `is_retriable()`, `is_not_found()`, `is_client_error()` and friends report the wrong answer. A cross-shard OCC abort is sent as `40001` (`serialization_failure`) precisely so the client retries it, and the client reports it as a non-retriable internal fault, so a retry loop abandons a transaction that was designed to be retried.

Client-side errors are unaffected, which localises the fault: connecting to a dead port still yields `SyncConnectionFailed` / `ErrorCode(3000)` / `retriable = true`. Only errors that arrive in a server frame are flattened.

### Steps to reproduce

```sql
-- Server: origin/main @ 28287607a, single node on loopback, fresh data dir,
-- config with [auth] mode = "trust". Driven through nodedb-client:
--     ConnectionBuilder::new(addr).username("admin").build()?
-- printing e.code(), e.is_retriable() and e.details() for each statement.

CREATE TABLE repro_t (id BIGINT PRIMARY KEY, v BIGINT);   -- OK
INSERT INTO repro_t (id, v) VALUES (1, 1);                -- OK

SELECT * FROM does_not_exist;
-- server sends SQLSTATE 42P01 "collection 'does_not_exist' not found"
-- client reports  ErrorCode(9000)  Internal { component: "unspecified", ... }
--                 <- WRONG, expected ErrorCode(1100) / CollectionNotFound

DROP TABLE does_not_exist;
-- server sends SQLSTATE 42P01 "collection 'does_not_exist' does not exist"
-- client reports  ErrorCode(9000)  Internal { ... }        <- WRONG

CREATE TABLE repro_t (id BIGINT PRIMARY KEY);
-- server sends SQLSTATE 42P07 "table 'repro_t' already exists"
-- client reports  ErrorCode(9000)  Internal { ... }        <- WRONG
```

The raw wire values above were read by logging `ErrorPayload.code` at the point the client currently discards it, so they are the bytes the server actually sent rather than an inference from the pgwire port.

### Expected behavior

A server error frame carries the server's own classification of the failure, so the client should reconstruct a typed `NodeDbError` from it: `42P01` should surface as `ErrorCode::COLLECTION_NOT_FOUND` with `ErrorDetails::CollectionNotFound`, `is_not_found()` true, and the server's message preserved. A SQLSTATE with no corresponding `ErrorDetails` variant should keep behaving exactly as it does today, as an internal error.

### Actual behavior

Every server-returned error, whatever its SQLSTATE, arrives as:

```
code      = ErrorCode(9000)
details   = Internal { component: "unspecified", detail: "<the real message>" }
retriable = false
```

## Root cause

The server computes a SQLSTATE (`nodedb/src/control/server/native/dispatch/conversion.rs`, `error_to_native`, plus `ddl_result_to_native` which forwards `DdlError.sqlstate` verbatim), and the wire type carries it with a doc comment naming this exact example:

```rust
// nodedb-types/src/protocol/frames.rs
pub struct ErrorPayload {
    /// SQLSTATE-style error code (e.g., "42P01" for undefined table).
    pub code: String,
    /// Human-readable error message.
    pub message: String,
}
```

Two functions in `nodedb-client/src/native/connection/mod.rs`, `check_error` and `response_to_query_result`, each do:

```rust
let msg = resp.error.map(|e| e.message)      // e.code discarded here
    .unwrap_or_else(|| "unknown error".into());
return Err(NodeDbError::internal(msg));      // everything becomes Internal
```

`.map(|e| e.message)` keeps the message and drops `e.code`. Nothing anywhere in the tree maps a SQLSTATE back to an `ErrorDetails`, so even a caller willing to do the work has no vocabulary to do it with, despite `nodedb-types/src/error/sqlstate.rs` already defining the constants for both directions.

Worth stating plainly, because it is easy to read this as intentional: a redaction measure would do the opposite of this. It would suppress the internal *message* and keep the coarse code. This keeps the full internal message, including raw parser output, and drops the harmless five-character classifier.

## How it survived

The test covering this branch constructs `NativeResponse::error(1, "42P01", "not found")` and asserts only that the *message* survives. Nothing asserts the code, so the missing mapping never failed a test.

## Provenance

Stated only because it explains the gap, not to attribute it:

- Introduced in `22d2c87ca` (2026-03-24), `feat(client): add native protocol client with connection pooling`, a single-parent commit on `main`. The repo's first PR (#1) was opened 2026-03-29, five days later, so this code was never reviewed.
- The server's SQLSTATE mapping predates it: `77d964688` (2026-03-17), a week earlier. `ErrorPayload.code` already carried its `42P01` doc comment when the client was written.
- The only later commit touching these lines is `25f6cea14`, a file-splitting refactor that moved them verbatim.

This looks like an oversight in an unreviewed initial implementation rather than a deliberate trade-off.

## Two adjacent findings, out of scope for this issue

Both were measured while reproducing the above. Filing separately if maintainers agree they are worth tracking.

1. **`error_to_native` types fewer errors than the pgwire path does.** Its match has arms for seven `crate::Error` variants and sends `XX000` for everything else, including `PlanError` and `UndefinedFunction`, whose own doc comments say the pgwire layer renders them as `42601` and `42883`. Measured on the same statements: `SELEKT 1` and `SELECT no_such_fn(1)` come back as `42601` and `42883` over pgwire but as `XX000` over the native protocol. So the native wire carries strictly less type information than pgwire for the same query, and fixing the client alone will not type those two classes.

2. **`42P07` has no `ErrorDetails` equivalent.** The server correctly sends `42P07` for a duplicate `CREATE TABLE`, but there is no "already exists" variant to map it onto, so it will keep falling back to `Internal` after the client fix.

Note also that a duplicate primary key is *not* an instance of this bug: the server itself sends `XX000` for it, because `RejectedConstraint` reaches `error_to_native`'s catch-all arm. The client folding that to `Internal` is correct.

### Proposed severity

### Reproducibility

Always — every attempt

### Environment & logs

macOS arm64 (Darwin 25.5.0), debug build of `origin/main @ 28287607a`.

Two things that otherwise cost time when reproducing:

- Trust auth is off by default, so a default-config server rejects the client's trust handshake. An `[auth]` section with `mode = "trust"` is needed, and `AuthConfig` has several fields with no serde defaults (`mode`, `superuser_name`, `min_password_length`, `max_failed_logins`, `lockout_duration_secs`, `idle_timeout_secs`, `max_connections_per_user`, `password_expiry_days`, `audit_retention_days`); the server reports them one at a time.
- Debug builds overflow the thread stack on some statements. `RUST_MIN_STACK=67108864` avoids it. This is a debug-build artifact, not a server defect.

### Before submitting

laksamanakeris:
Fixes #239.

## What was wrong

`check_error` and `response_to_query_result` in `nodedb-client/src/native/connection/mod.rs` both read a server error frame with `.map(|e| e.message)`, keeping the message and dropping `e.code`. Every server-returned error therefore reached callers as `ErrorCode(9000)` / `ErrorDetails::Internal`, no matter what SQLSTATE the server had computed for it.

Because every predicate on `NodeDbError` derives from `ErrorDetails`, this also broke classification: `is_retriable()`, `is_not_found()` and `is_client_error()` all reported the wrong answer. A cross-shard OCC abort is sent as `40001` specifically so the client retries it, and arrived as a non-retriable internal fault.

## What this changes

Adds `NodeDbError::from_sqlstate(sqlstate, message)` in `nodedb-types`, beside the SQLSTATE constants, so both transports can use it, and routes both client call sites through it.

It returns a whole `NodeDbError` rather than a bare `ErrorDetails`, which would be the more obvious shape, because `NodeDbError`'s fields are `pub(super)`: `nodedb-client` cannot assemble an error from parts, so a bare `ErrorDetails` would be unusable at the call site. Returning the whole error also keeps `code` and `details` in lockstep, which is the invariant every existing constructor maintains.

## Mapping policy

A SQLSTATE classifies a failure; it does not carry the operands. Mapping is therefore deliberately conservative, and the reasoning is recorded in the module doc rather than left implicit:

- **Payload fields are left empty, never guessed.** `42P01` becomes `CollectionNotFound { collection: "" }`. The failing name lives in the message, which is preserved verbatim; scraping it back out of prose would be brittle and would fabricate structured data. The variant, the `ErrorCode` and every derived predicate are exact.
- **Overloaded codes stay unmapped.** `55P03` is both `LOCK_NOT_AVAILABLE` (retriable) and `STALE_READ_NOT_LEADER` (not retriable); `57P03` and `53400` are likewise reused across two variants. Choosing either side would misreport retriability for the other.
- **Load-bearing payloads stay unmapped.** `57P04` needs `NotLeader { leader_addr }` and `54001` needs `FanOutExceeded { shards_touched, limit }`. Unlike a display name, a redirect address or a shard count is acted on, and an empty or zero value would be a fabricated instruction.
- **Codes with no variant stay unmapped**, for example `42P07` and `0A000`.

Everything unmapped falls through to `NodeDbError::internal`, which is exactly what every server error produced before. That is what makes the change non-regressive: a server emitting a SQLSTATE the client does not know yet is no worse off than today.

## Tests

- `42P01` yields `CollectionNotFound`, not `Internal`, with the message intact.
- An unmapped code (`XX000`) still yields `Internal`, asserted against `NodeDbError::internal` as the baseline. A deliberately-unmapped real code (`55P03`, `57P04`) takes the same path.
- Retriability is restored: `40001` and `57014` are retriable again; `42P01`, `42601`, `42883` and `42501` are asserted *not* retriable, since reporting a client error as retriable would make a caller spin.
- Payload fields are asserted empty rather than populated, so the fidelity limit is pinned by a test instead of only by a comment.
- The existing message-only test, which is what let this through, now asserts the code and the predicate as well. Both client call sites are covered, plus the no-payload fallback.

## Verified against a running server

Built `main` and drove the statements through `nodedb-client` on loopback. Before, all six error classes returned `ErrorCode(9000)`. After, `SELECT * FROM does_not_exist` and `DROP TABLE does_not_exist` return `ErrorCode(1100)` / `CollectionNotFound`. The transport control is unchanged: a dead port still gives `SyncConnectionFailed` / `ErrorCode(3000)` / retriable.

Two classes in that sample still return `Internal`, and correctly so given this change's scope. Both are recorded in #239 as separate findings: the server's `error_to_native` sends `XX000` for `PlanError` and `UndefinedFunction` over the native protocol even though the pgwire path types them as `42601` and `42883`, and `42P07` has no `ErrorDetails` equivalent to map onto.

## Gates

- `cargo fmt --all --check` clean.
- `cargo clippy -p nodedb-types -p nodedb-client --all-targets --all-features -- -D warnings` clean. The workspace-wide run fails in `nodedb-vector/src/quantize/pq.rs` (`nonminimal_bool`); confirmed identical on unmodified `main` with this local toolchain, so it is not from this change.
- `cargo nextest run` on the two changed crates: 787 pass, including the 13 added here.
- `cargo nextest run -p nodedb-client-tests --all-features`: 9 of 9 pass, including `native_execute_sql_with_bound_params_round_trips`, which drives the native protocol path both edited functions sit on. Needs `RUST_MIN_STACK` set, as CI already does.
- Workspace-wide, two failures occur on this machine that also occur on unmodified `main`, so neither comes from this change:
  - `nodedb-cluster-tests/tests/ilp_gateway_migration.rs` does not compile (`missing fields returning and rls_filters`). Excluded from the run.
  - `nodedb data::executor::handlers::join::shuffle_join::tests::shuffle_grace_infeasible_budget_is_deterministic_error` fails. Verified by checking out `28287607a` clean and running that single test, where it fails identically. It is unrelated to error mapping: it asserts `crate::Error::MemoryExhausted` from `drive_grace_build`, which is internal to the join executor and never touches `NodeDbError`.

  I did not get a clean workspace-wide `--no-fail-fast` run locally; two attempts were killed by `SIGTERM` during compilation. The one run that completed reached 4010 tests with the pre-existing failure above as the only one. Deferring to CI for the authoritative full-gate result.
- `cargo deny check` not run: `cargo-deny` is not installed in this environment. This change adds no dependencies, so its inputs are unchanged.

Also checked, since it is what would most plausibly break: nothing in `nodedb-client` or `nodedb-client-tests` asserts `is_internal`, `ErrorCode::INTERNAL`, `NDB-9000`, or `ErrorDetails::Internal`, so no existing test depended on the old flattening.

### DIFF
--- a/nodedb-client/src/native/connection/mod.rs
+++ b/nodedb-client/src/native/connection/mod.rs
@@ -13,9 +13,9 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use nodedb_types::error::{NodeDbError, NodeDbResult};
 use nodedb_types::protocol::{
     AuthMethod, CAP_COLUMNAR, CAP_CRDT, CAP_FTS, CAP_GRAPHRAG, CAP_SPATIAL, CAP_STREAMING,
-    CAP_TIMESERIES, FRAME_HEADER_LEN, HelloAckFrame, HelloFrame, Limits, MAX_FRAME_SIZE,
-    NativeRequest, NativeResponse, OpCode, PROTO_VERSION, RequestFields, ResponseStatus,
-    TextFields,
+    CAP_TIMESERIES, ErrorPayload, FRAME_HEADER_LEN, HelloAckFrame, HelloFrame, Limits,
+    MAX_FRAME_SIZE, NativeRequest, NativeResponse, OpCode, PROTO_VERSION, RequestFields,
+    ResponseStatus, TextFields,
 };
 use nodedb_types::result::QueryResult;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
@@ -449,24 +449,30 @@ fn io_err(e: std::io::Error) -> NodeDbError {
     NodeDbError::sync_connection_failed(format!("I/O: {e}"))
 }
 
+/// Rebuild the typed error a server error frame describes.
+///
+/// The frame carries the server's own SQLSTATE alongside the message; reading
+/// only the message would discard the one machine-matchable fact in it and
+/// leave every server-side failure indistinguishable from an internal fault.
+/// A frame with no payload at all has no classification to recover, so it
+/// falls back to `fallback` as an internal error.
+fn error_from_frame(error: Option<ErrorPayload>, fallback: &str) -> NodeDbError {
+    error.map_or_else(
+        || NodeDbError::internal(fallback),
+        |e| NodeDbError::from_sqlstate(&e.code, &e.message),
+    )
+}
+
 fn check_error(resp: NativeResponse) -> NodeDbResult<()> {
     if resp.status == ResponseStatus::Error {
-        let msg = resp
-            .error
-            .map(|e| e.message)
-            .unwrap_or_else(|| "unknown error".into());
-        return Err(NodeDbError::internal(msg));
+        return Err(error_from_frame(resp.error, "unknown error"));
     }
     Ok(())
 }
 
 fn response_to_query_result(resp: NativeResponse) -> NodeDbResult<QueryResult> {
     if resp.status == ResponseStatus::Error {
-        let msg = resp
-            .error
-            .map(|e| e.message)
-            .unwrap_or_else(|| "query failed".into());
-        return Err(NodeDbError::internal(msg));
+        return Err(error_from_frame(resp.error, "query failed"));
     }
     Ok(QueryResult {
         columns: resp.columns.unwrap_or_default(),
@@ -498,11 +504,67 @@ mod tests {
         assert_eq!(qr.rows[0][0].as_i64(), Some(42));
     }
 
+    /// An error frame must surface the server's classification, not just its
+    /// prose: asserting only on the message is what let the SQLSTATE be
+    /// dropped here unnoticed.
     #[test]
     fn response_to_query_result_error() {
         let resp = NativeResponse::error(1, "42P01", "not found");
         let err = response_to_query_result(resp).unwrap_err();
+
         assert!(format!("{err}").contains("not found"));
+        assert_eq!(
+            err.code(),
+            nodedb_types::error::ErrorCode::COLLECTION_NOT_FOUND
+        );
+        assert!(err.is_not_found());
+        assert!(!err.is_internal());
+    }
+
+    /// The same reconstruction happens on the non-query path, which has its
+    /// own copy of the error branch.
+    #[test]
+    fn check_error_reconstructs_sqlstate() {
+        let resp = NativeResponse::error(1, "42501", "authorization denied on t");
+        let err = check_error(resp).unwrap_err();
+
+        assert!(err.is_auth_denied());
+        assert_eq!(err.message(), "authorization denied on t");
+    }
+
+    /// Retriability is the behavioural consequence of the classification: a
+    /// serialization failure the client folds to `Internal` reads as
+    /// permanent, so a retrying caller gives up on a transaction that would
+    /// have succeeded.
+    #[test]
+    fn serialization_failure_frame_is_retriable() {
+        let resp = NativeResponse::error(1, "40001", "transaction aborted");
+        let err = response_to_query_result(resp).unwrap_err();
+
+        assert!(err.is_retriable());
+    }
+
+    /// Non-regression: a SQLSTATE with no mapped variant behaves exactly as
+    /// every server error did before the mapping existed.
+    #[test]
+    fn unmapped_sqlstate_frame_stays_internal() {
+        let resp = NativeResponse::error(1, "XX000", "boom");
+        let err = response_to_query_result(resp).unwrap_err();
+
+        assert!(err.is_internal());
+        assert!(format!("{err}").contains("boom"));
+    }
+
+    /// An error frame with no payload has no SQLSTATE to recover, so it keeps
+    /// the call site's own fallback text.
+    #[test]
+    fn error_frame_without_payload_uses_fallback() {
+        let mut resp = NativeResponse::error(1, "42P01", "not found");
+        resp.error = None;
+        let err = response_to_query_result(resp).unwrap_err();
+
+        assert!(err.is_internal());
+        assert!(format!("{err}").contains("query failed"));
     }
 
     #[test]
--- a/nodedb-types/src/error/ctors/from_sqlstate.rs
+++ b/nodedb-types/src/error/ctors/from_sqlstate.rs
@@ -1,0 +1,319 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Reconstruct a typed [`NodeDbError`] from a SQLSTATE received over the wire.
+//!
+//! The native protocol's `ErrorPayload` carries a five-character SQLSTATE and a
+//! human-readable message. The SQLSTATE is the server's own classification of
+//! the failure, so a client that reads only the message throws away every
+//! machine-matchable fact about the error and leaves callers with nothing to
+//! branch on.
+//!
+//! # Fidelity limits
+//!
+//! A SQLSTATE classifies; it does not carry operands. Variants with payload
+//! fields (`CollectionNotFound { collection }`, `UndefinedFunction { name }`)
+//! are therefore rebuilt with those fields empty: the failing name lives in the
+//! message, which is preserved verbatim, but it is not recoverable as
+//! structured data. The variant, the [`ErrorCode`], and every predicate derived
+//! from them ([`NodeDbError::is_retriable`], [`NodeDbError::is_not_found`], …)
+//! are exact. An empty field reads as "not transmitted"; guessing the operand
+//! out of the message text would be worse than leaving it blank.
+//!
+//! Codes are mapped only where a SQLSTATE and an [`ErrorDetails`] variant
+//! correspond one-to-one. Three groups are deliberately left unmapped:
+//!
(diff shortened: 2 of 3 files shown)


laksamanakeris:
Following up on the test note in the description.

**`nodedb-client-tests`, the crate that exercises this change end to end: 9/9 pass**, including `native_execute_sql_with_bound_params_round_trips`, which drives the native protocol path both edited functions sit on.

That run needs `RUST_MIN_STACK=67108864`. Without it, `native_execute_sql_params_round_trip` dies with `SIGABRT` / "has overflowed its stack" before reaching any assertion. It is the known debug-build stack artifact, not a behavioural failure, and it reproduces independently of this change.

**Changed crates: 787/787 pass**, including the 13 tests added here.

I was not able to finish a clean workspace-wide `--no-fail-fast` run locally. Two attempts were killed by `SIGTERM` during compilation, and the recompile is long enough that retrying was not a good use of time. What I do have:

- One workspace run completed and reached 4010 tests before fail-fast stopped it, with exactly one failure: `nodedb data::executor::handlers::join::shuffle_join::tests::shuffle_grace_infeasible_budget_is_deterministic_error`.
- That failure is pre-existing. Verified by checking out `28287607a` clean and running that single test, where it fails identically. It asserts `crate::Error::MemoryExhausted` out of `drive_grace_build`, which is internal to the join executor and never touches `NodeDbError`.
- `nodedb-cluster-tests/tests/ilp_gateway_migration.rs` does not compile on `main` either (`missing fields returning and rls_filters`), so it was excluded from the run.

One further check worth recording, since it is the thing that would most plausibly break: nothing in `nodedb-client` or `nodedb-client-tests` asserts `is_internal`, `ErrorCode::INTERNAL`, `NDB-9000`, or `ErrorDetails::Internal`. No existing test depended on the old flattening behaviour, so there is no implicit contract this change violates.

Deferring to CI for the authoritative full-gate result.

laksamanakeris:
Closing: superseded by `5ef7fc13c` (`feat(client): propagate typed error classification over the native wire`), which fixed the same bug on `main` three days after this PR was opened, and did it better.

This PR reconstructed the classification on the client by mapping SQLSTATE back to `ErrorDetails`. `5ef7fc13c` widened the channel instead: `ErrorPayload` gained `ndb_code`, the stable numeric NodeDB code, and `NodeDbError::from_wire` rebuilds the typed error from that. Since SQLSTATE is many-to-one, the number is the only thing that can carry the classification, so the reverse mapping here is not just redundant, it is the design `from_wire`'s module doc rules out. Both files this PR edits have been rewritten since, and `git merge-tree` conflicts in both.

The part of #239 that is still open is different in kind and needs a different fix: native frames authored as a bare SQLSTATE, with no classified `Error` behind them, still ship `ndb_code == 0`, so DDL refusals (`DROP TABLE` naming an absent collection, and every other `DdlError`) and the Control-Plane guard rejections continue to reach the client as `Internal`. Details re-measured in https://github.com/NodeDB-Lab/nodedb/issues/239#issuecomment-5294773345; a replacement PR follows.

PR #243, opened 2026-08-23
fix(swim): fast-restart rejoin — persist incarnation, echo refutations, ping liveness

### LINKED ISSUE #165: [Epic] P2 — Cluster Consensus Safety (v0.6)
**Epic — v0.6 production-hardening, phase P2: Cluster Consensus Safety**

Goal: a restart or partition never produces two leaders, divergent committed entries, or silent state-machine divergence; linearizable reads are actually linearizable.

Every box below is checked against this repository's `main`. 

### Critical (filed)
- [x] #161 — Raft HardState never persisted (+ reply-before-durable ordering) → split-brain on restart. Persist before votes and replies leave the node, 9469296c2.
- [x] #162 — data-group InstallSnapshot advances the index without the data → silent divergence. `install_snapshot/finalize.rs` applies to the state machine before advancing the Raft boundary. An apply failure returns without advancing, so the follower retries.

### High
- [ ] Deposed/partitioned leader serves stale linearizable reads — no ReadIndex heartbeat, no leader lease. **Partial.** ReadIndex closes the safety hole: `start_read_index` issues its own quorum round and returns `ReadIndexStatus`, which separates "not yet confirmed" from "leadership lost". f9910444b, b1f52e9a4. Dead `follower_read.rs` deleted. Check-quorum has since landed: a leader tracks when a majority last acknowledged it and demotes itself once that gap reaches an election timeout, `nodedb-raft/src/node/quorum_contact.rs`, refreshed from `rpc/append_entries.rs:145`. Remaining: leader lease, a latency optimization over the ReadIndex round, not safety.
- [x] BoundedStaleness freshness measures time-since-last-apply, not lag-vs-leader — a badly-lagging follower reports "fresh". `nodedb-raft/src/node/staleness.rs` returns `Behind` when `last_applied` trails the leader's commit index. b1f52e9a4. Dead `closed_timestamp.rs` deleted.
- [ ] Rolling upgrade blocked on any wire bump (`MIN_WIRE_FORMAT_VERSION == WIRE_FORMAT_VERSION`) → version hard-partition mid-upgrade. **Not started here.** `nodedb-types/src/wire_version.rs:52` still reads `MIN_WIRE_FORMAT_VERSION = WIRE_FORMAT_VERSION`.
- [x] No fencing tokens anywhere in raft/cluster. Two layers land. Transport: `cluster_epoch` is stamped in the RPC header and enforced at decode, `rpc_codec/header.rs`. Catalog: descriptor-version fencing on every dispatch path, 6023d2aa0, 5445f8bf9, a18c74b38 — see **Descriptor-version fencing**.

### Medium / Low
- [x] Scatter-gather has no shard-failure/timeout guard → hang or silently-partial aggregate. Four coordinators rebuilt on one shape: typed error, proof-carrying totals with private fields, `BTreeSet` responded-tracking, 30s deadline. b037bab06, ae7718f52, 516d263cd.
- [ ] Descriptor-lease expiry uses raw cross-node wall clock, no skew bound / fencing token. **Partial.** The drain no longer judges another node's deadline by its own clock; an explicit `DescriptorDrainEnd` is the only thing that clears a drain. 1f67bb414. The skew bound the clamp derives from now exists: `MAX_CLOCK_SKEW_NS` (5s) with `HlcClock::update_checked` refusing a remote HLC further ahead, `nodedb-types/src/hlc.rs:25`, wired at metadata-apply ingress. bb7a94911. Remaining: apply that clamp to lease expiry itself, and a SWIM-Dead to `on_node_crash` release hook.
- [x] SWIM fast-restart rejoin can stick (restart at incarnation 0 refuted vs lingering Dead). The incarnation is persisted and bumped on restart, `nodedb-cluster/src/bootstrap/start.rs:84` — a node that crashed while peers held `Dead(A, N)` restarts at `N + 1`, so its first `Alive` dominates every lingering rumour. A read error fails bootstrap rather than silently restarting at 0.
- [x] No pre-vote (partition-heal term inflation forces healthy leader to step down); no leadership transfer (`TimeoutNow`). Pre-vote grants require Follower or Candidate role, a higher term, the up-to-date log rule, and time-bounded leader stickiness. dd91eed70. TimeoutNow, ae45049bf.
- [x] Snapshot GC `sweep_orphans` runs at startup only; decommission doesn't propagate cluster-wide ShutdownWatch. GC also runs every ~60s from the Raft tick loop, `raft_loop/loop_core.rs:418`. Decommission signals through consensus, 0f625a6a2, 7056676cc.

### State-machine divergence

- [x] A committed entry could apply twice. One Raft batch carried the same committed index twice, and `apply_committed` forwarded both copies. A PK-carrying write absorbs the repeat as an upsert and hides it; an append-shaped write (timeseries/columnar/spatial insert) gains a row. Each committed index is now handed off at most once, and the delivery watermark is claimed in one critical section — reading it and advancing it under separate lock acquisitions let two concurrent deliveries claim the same range. ea4f73fa4. Reproduced 3-in-6 before, 10/10 clean after.
- [ ] Why the Raft layer emits a batch containing a repeated committed index is unexplained. The applier enforces its own contract, which is the right boundary and makes the duplicate unreachable, but the upstream duplication may be a defect in log delivery. Start at `nodedb-cluster` raft loop batch construction.
- [x] Raft readiness was bounded by a fixed 30s wall clock, so a node with a large metadata replay was killed mid-boot while healthy. The gate now bounds on lack of progress: the deadline resets whenever `applied_index` advances, and a stuck group still fails, naming the index it stuck at. #253, 0e1f6a3d4.
- [x] `HlcClock::update` had no caller at any node boundary and no skew bound, so cross-node HLC comparisons carried an unbounded offset and one far-future observation moved a node permanently. `update_checked` refuses a remote HLC more than `MAX_CLOCK_SKEW_NS` ahead; the one remote observation the metadata group carries is folded at apply, and every other `Hlc` on a `MetadataEntry` is a future deadline that must not be. #264, bb7a94911.

---

## Descriptor-version fencing

A descriptor lease does not prove a plan is current. `force_refresh_lease` never compares the requested version against the catalog. A plan stamped just before a DDL commits still acquires its lease afterwards, at the version that DDL superseded.

Two dispatch paths carried a plan across that window.

| Path | Window | Commit |
|---|---|---|
| Interactive transaction | Statement time to COMMIT. Bounded only by how long the client holds the block open. | a18c74b38 |
| DEFINE EVENT THEN action | Plan to lease acquisition, then across each task's WAL append and SPSC round trip. | 5445f8bf9 |

`gateway::version_check::check_descriptor_holds` owns the comparison. Holds are deduplicated by scope identity, then grouped by their own `(database, tenant)`. A collection name is never compared against another tenant's catalog. Non-collection descriptor kinds are skipped. A stale hold aborts COMMIT with SQLSTATE `40001` before any durable write.

The event-action fence re-runs per dispatch. One check before the loop clears task 0, then lets a later task run against a version the catalog has left behind.

**Correction.** Two shipped module comments claimed a timed-out drain "is force-ended and the DDL proceeds anyway". It does not. The timeout proposes `DescriptorDrainEnd` to unblock new acquires, then `drain_for_ddl` returns `Err` and the DDL aborts. The fences are still required. Their stated reason was wrong. Both comments now name the real window above, and the real carve-out: a mixed-version cluster skips the drain outright.

## Deferred actions

Auditing the trigger fence surfaced three defects in the Event Plane retry path. All three were confirmed in code before any edit.

- [x] The retry re-fired triggers that had already succeeded. `dispatcher/single.rs` called `fire_for_operation` with no trigger filter. The failed trigger's name was stored and never read.
- [x] A failing trigger cancelled its siblings. `fire_common.rs` returned `Err` on the first failure, so triggers queued behind it never ran on any path.
- [x] DEFINE EVENT failures had no retry path. `consumer_helpers.rs` passed no queue.
- [x] Retry unit is one action, not one source write. Records are keyed for idempotent enqueue and backed by a per-core redb store that opens lazily. 02f8a20c5.
- [x] Actions commit as one transaction, `control/system_txn/`. Partial dispatch became unrepresentable, so that variant and its per-dispatch fence are deleted — the COMMIT fence covers them.
- [x] DLQ is readable and replayable. `SHOW TRIGGER DLQ` and `REQUEUE TRIGGER DLQ <entry_id>`. It was durable but write-only, with no payload to replay. ddd8d0921.
- [x] Event-action writes no longer cascade without bound. They dispatched as `EventSource::User`, and the event-definition path gated on `is_data_event()` but never on the source. An action writing to the collection it watches re-fired itself forever.
- [ ] Trigger bodies are procedural blocks, not transactions. A body that fails mid-block has partially applied, and its retry repeats what landed. Apply the `system_txn` treatment event actions already get.

## Test infrastructure

- [x] Both `crash_resp_kv_write` tests join the `load-sensitive` nextest group. Each fails the harness 20s readiness poll under the full suite and passes in ~5s in isolation. The child server is starved of cores by the unit-test fan-out, not slow. 805ebde55.
- [x] `calvin_multishard_write_in_explicit_block_commits` failed 3 of 3 tries on "no sequencer leader elected yet". Its gate waited for `sequencer_metrics` to be set, which proves only that `start_raft` wired the service. It now waits on `raft_status_fn` reporting a `SEQUENCER_GROUP_ID` leader. de48fba51.

EnRaiha:
Closes the SWIM fast-restart rejoin stick (epic #165, Medium #9): a node that crashes while peers hold Dead(A, N) restarts at incarnation 0, announces Alive(0) exactly once, and can diverge forever when the probabilistic refutation round-trip fails.

- F4 (primary): persist local incarnation in the catalog (KEY_SWIM_INCARNATION); bootstrap resumes at persisted+1 so the first announcement dominates any lingering Dead rumour; every self-refutation bump persisted fire-and-forget.
- F1: refutations echoed deterministically on Ack + forwarded PingReq piggyback.
- F2: a ping IS liveness evidence — sender Alive claim applied before ack; self-advertise rate-limited to 500ms.
- F3: fresh Alive apply cancels the pending suspicion timer (kills the re-kill race).

6 regression tests incl. deterministic recovery loop (A holds Dead(B,5) → echo → self-refute → converge Alive). Verification: swim 129/129, cluster lib 1023, clippy -D warnings 0, maya-gate L1 clean.

Closes #165 item: Medium — SWIM fast-restart rejoin.

### DIFF
--- a/nodedb-cluster/src/bootstrap/start.rs
+++ b/nodedb-cluster/src/bootstrap/start.rs
@@ -78,8 +78,10 @@ pub fn register_default_subsystems(
     // incarnation. A node that crashed while peers held `Dead(A, N)`
     // restarts at `N + 1`, so its very first `Alive` announcement
     // dominates every lingering rumour — no probabilistic refutation
-    // round-trip required.
-    let persisted_inc = catalog.load_swim_incarnation().unwrap_or(None);
+    // round-trip required. A read error (or corrupted metadata) FAILS
+    // bootstrap rather than silently restarting at incarnation 0, which
+    // would reintroduce the stick the persisted key exists to prevent.
+    let persisted_inc = catalog.load_swim_incarnation()?;
     let initial_incarnation = match persisted_inc {
         Some(v) => Incarnation::new(v).bump(),
         None => Incarnation::ZERO,
--- a/nodedb-cluster/src/bootstrap/start.rs
+++ b/nodedb-cluster/src/bootstrap/start.rs
@@ -34,6 +34,9 @@ use crate::subsystem::{
     DecommissionSubsystem, ReachabilitySubsystem, RebalancerSubsystem, RunningCluster,
     SubsystemRegistry, SwimSubsystem, SwimSubsystemConfig,
 };
+use crate::swim::config::SwimConfig;
+use crate::swim::incarnation::Incarnation;
+use crate::swim::incarnation_store::CatalogIncarnationStore;
 use crate::transport::NexarTransport;
 
 use super::bootstrap_fn::bootstrap;
@@ -69,9 +72,24 @@ pub fn register_default_subsystems(
     config: &ClusterConfig,
     ctx: &BootstrapCtx,
     executor: Arc<MigrationExecutor>,
+    catalog: &Arc<ClusterCatalog>,
 ) -> crate::error::Result<()> {
+    // Fast-restart rejoin safety: resume SWIM above the last persisted
+    // incarnation. A node that crashed while peers held `Dead(A, N)`
+    // restarts at `N + 1`, so its very first `Alive` announcement
+    // dominates every lingering rumour — no probabilistic refutation
+    // round-trip required.
+    let persisted_inc = catalog.load_swim_incarnation().unwrap_or(None);
+    let initial_incarnation = match persisted_inc {
+        Some(v) => Incarnation::new(v).bump(),
+        None => Incarnation::ZERO,
+    };
+
     let swim_cfg = SwimSubsystemConfig {
-        swim: crate::swim::config::SwimConfig::default(),
+        swim: SwimConfig {
+            initial_incarnation,
+            ..SwimConfig::default()
+        },
         local_id: NodeId::try_new(config.node_id.to_string()).map_err(|e| {
             crate::error::ClusterError::Config {
                 detail: format!("node_id is not a valid ID: {e}"),
@@ -85,6 +103,7 @@ pub fn register_default_subsystems(
             a
         }),
         seeds: config.seed_nodes.clone(),
+        incarnation_store: Some(Arc::new(CatalogIncarnationStore::new(Arc::clone(catalog)))),
     };
 
     registry.register(Arc::new(SwimSubsystem::new(
@@ -186,6 +205,7 @@ pub async fn start_cluster_subsystems(
     routing: Arc<std::sync::RwLock<crate::routing::RoutingTable>>,
     transport: Arc<NexarTransport>,
     raft_multi_raft: Arc<std::sync::Mutex<crate::multi_raft::MultiRaft>>,
+    catalog: &Arc<ClusterCatalog>,
 ) -> Result<RunningCluster> {
     let health = ClusterHealth::new();
     let (decommission_signal, _) = tokio::sync::watch::channel(false);
@@ -206,7 +226,7 @@ pub async fn start_cluster_subsystems(
     ));
 
     let mut registry = SubsystemRegistry::new();
-    register_default_subsystems(&mut registry, config, &ctx, executor)?;
+    register_default_subsystems(&mut registry, config, &ctx, executor, catalog)?;
 
     registry
         .start_all(&ctx)
--- a/nodedb-cluster/src/catalog/core.rs
+++ b/nodedb-cluster/src/catalog/core.rs
@@ -175,7 +175,15 @@ impl ClusterCatalog {
                     arr.copy_from_slice(bytes);
                     Ok(Some(u64::from_le_bytes(arr)))
                 } else {
-                    Ok(None)
+                    // Corrupted metadata: an unexpected length is NOT "no
+                    // value". Surface it so operators detect catalog
+                    // corruption early instead of silently restarting at
+                    // incarnation 0 (which would reintroduce the rejoin
+                    // stick the key exists to prevent).
+                    Err(catalog_err(format!(
+                        "metadata key {KEY_SWIM_INCARNATION} has unexpected length {} (expected 8)",
+                        bytes.len()
+                    )))
                 }
             }
             None => Ok(None),
--- a/nodedb-cluster/src/catalog/core.rs
+++ b/nodedb-cluster/src/catalog/core.rs
@@ -17,8 +17,8 @@ use crate::error::Result;
 use super::migration::migrate_if_needed;
 use super::schema::{
     CATALOG_FORMAT_VERSION, GHOST_TABLE, KEY_CA_CERT, KEY_CLUSTER_EPOCH, KEY_CLUSTER_ID,
-    KEY_FORMAT_VERSION, METADATA_TABLE, MIGRATION_STATE_TABLE, ROUTING_TABLE, TOPOLOGY_TABLE,
-    catalog_err,
+    KEY_FORMAT_VERSION, KEY_SWIM_INCARNATION, METADATA_TABLE, MIGRATION_STATE_TABLE, ROUTING_TABLE,
+    TOPOLOGY_TABLE, catalog_err,
 };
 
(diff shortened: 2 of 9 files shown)


EnRaiha:
Addressed the two critical review items in 1876085ad:
1. `register_default_subsystems` now propagates catalog read errors (`?`) instead of `unwrap_or(None)` — bootstrap fails on IO/corruption instead of silently restarting at incarnation 0.
2. `load_swim_incarnation` now returns an error (with key + length) when stored metadata length != 8 — corrupted catalogs surface early.

Thanks for the review!

farhan-syah:
The diagnosis was right, and the fix is actually stronger than the description claims. `ProbeScheduler::reshuffle` builds the probe queue from `snapshot().alive()` only, so a node marked Dead is never probed again — peers holding `Dead(A, N)` send A nothing at all. Once the rumour ages out of piggyback buffers there is no path left for A to learn it is considered dead, so this is not just a probabilistic stick, it is a permanent one-way split.

F1 and F2 together are the part I would highlight: A pings with its stale Alive, the peer refutes and echoes its stored Dead record on the reply, A self-refutes above it, converged. That is a deterministic recovery loop and it works without persistence at all — F4 just makes the first announcement win outright. I checked whether F2 lets a stale claim bypass incarnation ordering and it does not: the sender's claim goes through `apply_and_notify` like any other update.

Landed in 47f890dd9, with you as co-author.

Four corrections went in on top, as 40ea609a8 and one hunk in the first commit.

1. `save_swim_incarnation` was a blind overwrite. Bumps are persisted off the probe path, so two of them can reach the catalog out of order and leave the lower value on disk. The node then restarts at an incarnation it has already advertised and loses to a peer still holding a Dead record at that number — the exact stick the key exists to break. The write now reads inside its own transaction and discards anything at or below what is stored. `swim_incarnation_never_moves_backwards` covers it.

2. The persistence ran in `tokio::spawn`, but `save` is a synchronous redb commit — a blocking call on a reactor worker. Now `spawn_blocking`. Note these two interlock: taking the write off the probe path is what makes bumps land out of order, so it is only safe because the store is monotonic now.

3. `spawn_with_subscribers` gained a parameter but `nodedb-cluster-tests/tests/swim_routing_invalidation.rs` was left on the old arity, so the branch does not compile against the integration suite.

4. `CatalogIncarnationStore` held `Arc<ClusterCatalog>`. The failure detector outlives the shutdown sequence that closes the catalog, so the redb file lock survived it and the next open failed with "Database already open. Cannot acquire lock." A node could not restart against its own data directory — four restart tests fail on it. It now holds a `Weak`, matching what `LeaseRenewalLoop` already does with `Weak<SharedState>` for the same reason.

Items 3 and 4 both surface immediately in the second test stage (`-p nodedb-cluster-tests`), which the verification note does not mention — `swim 129/129` and `cluster lib 1023` are both first-stage. Worth running both stages before posting results; item 4 in particular is a real availability bug, not a test artifact.

PR #244, opened 2026-08-23
fix(wire): open [MIN,WIRE] window — rolling upgrades possible

EnRaiha:
Fixes epic #165 High #5 (wire version hard-partition mid-upgrade). Replaces the exact-equality join gate with a range window [min_wire_version, CLUSTER_WIRE_FORMAT_VERSION] so mixed-version clusters can form (N-1 rolling upgrade).

- MIN_WIRE_FORMAT_VERSION=1, WIRE_FORMAT_VERSION=2; doc rewritten for window semantics (schema vs transport explicitly separate).
- wire_version_in_window() pure gate + effective floor (max(compile-time MIN, persisted ClusterSettings.min_wire_version)) — operator knob now enforced.
- ClusterVersionView moved from nodedb (inverted dep) to nodedb-cluster/src/topology/version_view.rs; re-export shim; dangling rustdoc links fixed.
- restart re-stamps self NodeInfo.wire_version after upgrade (raise-only, idempotent; pure helper, unit-tested without a live transport).
- window_from_join_response() logs observed cluster version window.

Verification: nodedb-types+nodedb 6246 pass, cluster lib 1023/1023, clippy 0, zero exact-equality gates left, maya-gate clean 12/12. Note: 32 rustdoc broken-intra-doc-link errors are PRE-EXISTING in untouched files (tracked separately).

Part of #165.


farhan-syah:
Closing this one without taking the change, and the reason is different from the other four — the premise rather than the implementation.

`nodedb-types/src/wire_version.rs` carries a "DO NOT BUMP THIS BEFORE 1.0" section that sets out why the floor and ceiling are deliberately equal, and it anticipates this change specifically: "After 1.0, when real deployments exist and a genuine compatibility window is introduced, this becomes meaningful — bump it then, deliberately, and only alongside an actual `MIN_WIRE_FORMAT_VERSION < WIRE_FORMAT_VERSION` support window." That is precisely this PR, at precisely the point the doc says to hold off.

The governing fact is that there are no deployed clusters before 1.0, so there is no older peer a new build has to talk to. A hard partition mid-upgrade cannot happen when there is nothing to upgrade from. The doc also notes the consequence: while floor equals ceiling, every `wire_version >= V` feature gate is unreachable, because inside a cluster that exists all nodes are provably on the same version.

There is also a sequencing problem independent of the pre-1.0 policy. The change bumps `WIRE_FORMAT_VERSION` to 2 while opening the window to [1, 2], but nodes already running the current code are at 1 and still gate on exact equality. So a v2 node joining an existing v1 cluster is rejected by the v1 side, while a v1 node joining a v2 cluster is accepted. A rolling upgrade means upgrading nodes one at a time inside an existing cluster, which is the rejected direction. The window can only take effect once every node already runs window-capable code, so the order has to be: ship the window logic at 1/1 with no behavioural change, let it reach every node, then bump. Doing both together is the one ordering that cannot deliver the feature.

On the `ClusterVersionView` move: there is no inverted dependency to fix. The type lives in `nodedb` and is used only within `nodedb`; `nodedb-cluster` does not depend on `nodedb`. Moving a host-layer concept down into the cluster crate would make the layering worse rather than better.

One observation from the PR was right and has been fixed separately: `ClusterSettings.min_wire_version` is documented as a minimum peers must speak, and nothing enforces it — it appears only in its own definition and tests. Rather than wire it up, its doc now states plainly that it is reserved and inert while floor equals ceiling, and what it will mean once a window opens. An operator knob that silently does nothing is worth correcting on its own.

When the window is genuinely wanted after 1.0, the `wire_version_in_window` gate and the effective-floor rule from this branch are a reasonable starting point — the objection is to the timing and the simultaneous bump, not to that logic.

PR #245, opened 2026-08-23
fix(cluster): enforce cluster-epoch fence at the decode chokepoint

EnRaiha:
Makes the cluster-epoch fence real (epic #165 High #6). cluster_epoch was stamp-only: docs claimed stale-epoch peers are rejected but no rejection existed.

- ClusterError::StalePeerEpoch { peer_epoch, local_epoch } for operator-visible diagnostics.
- validate_peer_cluster_epoch(rpc_type, peer_epoch): rejects peer_epoch < local unless exempt (EPOCH_EXEMPT_RPC_TYPES = JOIN_REQ/JOIN_RESP/PING/PONG — join is how a fenced peer re-adopts the epoch; ping/pong is the pre-join liveness channel).
- Single enforcement point in parse_frame (the only decode path, both directions) before payload/CRC work; no dispatch/transport changes.
- EPOCH_TEST_LOCK serialises every test mutating the shared epoch global (fixed a latent parallel race in two pre-existing header tests).

9 new tests; cluster_epoch 11/11 + header 9/9 stable across 5 runs, full lib 1016/1016 serial, clippy 0, maya-gate clean 4/4.

Part of #165.

### DIFF
--- a/nodedb-cluster/src/cluster_epoch.rs
+++ b/nodedb-cluster/src/cluster_epoch.rs
@@ -42,7 +42,8 @@ use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
 
 use crate::catalog::ClusterCatalog;
-use crate::error::Result;
+use crate::error::{ClusterError, Result};
+use crate::rpc_codec::discriminants::{RPC_JOIN_REQ, RPC_JOIN_RESP, RPC_PING, RPC_PONG};
 
 /// Process-global cluster epoch high-water mark.
 ///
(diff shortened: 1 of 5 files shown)


EnRaiha:
Addressed the critical transport-handling item in e9b1a46b8: `handle_stream` now matches `ClusterError::StalePeerEpoch` explicitly — the frame is dropped with a warn! log (fenced until rejoin) and the stream task returns Ok, so the connection stays open for exempt join/ping traffic instead of ending the stream via generic error propagation. The `#[error]` attribute formatting is confirmed by clippy -D warnings 0.

farhan-syah:
You were right that the fence was missing. `cluster_epoch.rs` documented rejecting frames "from peers stuck on a strictly older epoch" and no rejection existed — I'd assumed it did until I read it against your PR. Thanks for that.

Closing this because the fix went underneath it rather than on top. The reason is that `observe_peer_cluster_epoch` did `fetch_max` on every inbound frame, so a node's epoch meant "highest number I have seen on a wire", not "generation whose state I have applied". Rejecting on that value inverts the check: a node that receives one frame stamped 9 starts refusing peers stamped 8 while its own routing table may still hold generation 7. It fences others on a number it overheard, while itself acting on stale state.

The exemption list and the rejection window after every metadata leader election are downstream of that — the fence firing on gossip propagation order rather than on an agreed fact.

What landed instead, in 833a6014f and e16d7420c:

- `applied` and `observed` are tracked apart. Observing a higher epoch is evidence this node is behind; it never advances what the node claims.
- The epoch is a committed `MetadataEntry::ClusterEpochBump`. The metadata leader proposes it and every node advances by applying the same entry, so the generation has an agreed value and an order. Previously it was a local counter incremented as a side effect of noticing leadership, with no consensus behind it at all.
- A node fences itself, not its peers: `observed > applied` means it missed a transition, so it stands down from coordinating queries until the metadata group delivers the bump. Raft traffic is never gated, since that is how it catches up. This removes the need for the exemption list and the silent-drop response path.
- The epoch moved off the process-global atomic onto the transport's `AuthContext`. Nodes sharing a process no longer alias into one epoch.

Two notes from your PR worth recording. The `EPOCH_TEST_LOCK` race you found between the `header` and `cluster_epoch` test modules was real, and the restructure removes the shared global it depended on. And the module doc's claim that threading the epoch would touch "hundreds of sites" was wrong — it was 18 call sites in 3 files.

PR #246, opened 2026-08-23
fix(lease): descriptor-lease crash-wedge GC

EnRaiha:
Fixes epic #165 Medium #8 availability gap: a node that crashes while holding a descriptor lease leaves its records in every MetadataCache forever — every DDL on those descriptors wedges on the 35s drain wait and fails permanently (shutdown_release's 'leases drain via TTL' doc claim was wrong; the poller never checked expiry).

- Drain filter: count_matching_leases ignores leases whose holder is no longer a cluster member (fail-safe: missing topology = all members) and leases past expires_at (matches grant-path semantics).
- TopologyChange::Leave hook: spawn a leader-gated (is_singleton_worker) task proposing DescriptorLeaseRelease for the departed node (spawned, not inline — apply runs on the raft-loop task; inline propose-wait would deadlock the applied watcher).
- Periodic sweep (leader-gated, tick-throttled, mirrors reconcile_placement): proposes releases for every holder no longer in ClusterTopology — safety net for crashes before Leave commits.

Tests: 78 lease (3 drain-filter + 3 gc), 3 lease_gc collector; clippy 0, check 0 errors, maya-gate clean 11/11.

Part of #165.

### DIFF
--- a/nodedb-cluster/src/raft_loop/builder.rs
+++ b/nodedb-cluster/src/raft_loop/builder.rs
@@ -8,13 +8,14 @@
 //! than in [`super::loop_core`] so that the struct definition, constructor,
 //! and runtime methods stay under the file-size limit.
 
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
 use crate::applied_watcher::GroupAppliedWatchers;
 use crate::catalog::ClusterCatalog;
 use crate::forward::PlanExecutor;
 use crate::metadata_group::applier::MetadataApplier;
+use crate::metadata_group::cache::MetadataCache;
 
 use super::hooks::{
     AssignRemoteSurrogate, CalvinSubmit, CalvinSubmitInbox, ReleaseReservation, ReserveRead,
@@ -65,6 +66,7 @@ impl<A: CommitApplier, P: PlanExecutor> RaftLoop<A, P> {
             // Construction-time builder, before `run()` and any join kick — a
             // fresh `Notify` has no pending permit, so this loses nothing.
             reconcile_notify: tokio::sync::Notify::new(),
+            metadata_cache: self.metadata_cache,
         }
     }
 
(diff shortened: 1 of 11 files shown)


EnRaiha:
Addressed review items in 61c184bdd: proposal failures now log at warn (operators can notice GC not happening) and the drain-filter doc comment is fully English. Thanks for the review!

farhan-syah:
Good find, and the diagnosis was right on both halves. `count_matching_leases` filtered only on descriptor id and version — no expiry, no membership — and `shutdown_release` logs "leases will drain via TTL" on both its failure paths while nothing anywhere reads `expires_at`. Those leases never drained, and the DDL wedge was permanent.

The three-part shape is right too, and I kept it: the drain filter for liveness, the Leave hook for immediate reclaim, the sweep as the safety net. Spawning the Leave hook instead of running it inline is correct — an inline propose-and-wait would deadlock the applied-index watcher.

Landed in 06b51a2e4, with you as co-author.

One correction went in on top, as 7168ecc55. The filter compared `l.expires_at > shared.hlc_clock.peek()`. `HlcClock::peek` returns the last HLC the clock observed and never advances on its own, so on a quiet node it sits far behind wall time — at `Hlc::ZERO` when nothing has stamped it. Every lease then compares as unexpired and the filter does nothing. That is exactly the case the PR targets: a crashed node on an idle cluster, where its leases are the only ones left and nothing is generating HLC events. The comparison now uses `wall_now_ns()` against `expires_at.wall_ns`, which is what `renewal.rs` and the drain TTL already do, and `renewal::tick` documents why.

The tests could not catch it because `expired()` and `unexpired()` derived their timestamps from `hlc_clock.peek()` as well, so the fixture and the code under test shared one frozen frame and the assertions held regardless. `expired()` computed `0.saturating_sub(1_000)`, which is `0`, and `Hlc(0,0) > Hlc(0,0)` is false — so it passed for a reason unrelated to expiry. Both helpers now use real wall time, and two tests were added: one that asserts the HLC is unadvanced and still expects the expired lease to be dropped, and one for the opposite direction, where an HLC dragged ahead by a peer with a skewed clock must not make a live lease look expired. Dropping a live hold would trade the wedge for a correctness bug.

Also worth recording: the expiry filter is safe only because `lease/renewal.rs` re-acquires any lease approaching expiry, so an expired record means that holder's renewal stopped. A live hold on the draining node itself stays counted through `lease_refcount`, which the filter leaves alone.

Issue #162, opened 2026-07-05
Data-group InstallSnapshot advances the Raft index without transferring state-machine data — silent divergence


habibtalib:
**Source:** static engine audit, 2026-07-05 (cluster/consensus lens), verified against `main @ d92ede42`
**Severity:** Critical — silent state-machine divergence on a lagging follower.

## Summary
Data-group `InstallSnapshot` fast-forwards Raft's `commit_index`/`last_applied` to `last_included_index` and compacts the log, but transfers **no state-machine data** — the sender hardcodes `snapshot_bytes: &[]` and the receiver ignores `req.data`. The assembled `.snap` file is written but **never read back** (the only `.snap` reference is the write site). The engine never *produces* real snapshot bytes, so the whole snapshot-install path is a non-functional stub for data groups. The code comments the hole directly.

## Location
- `nodedb-cluster/src/raft_loop/tick.rs:290` (`snapshot_bytes: &[]`)
- `nodedb-raft/src/node/rpc/install_snapshot.rs:34-52` (ignores `req.data`)
- `nodedb-cluster/src/raft_loop/handle_rpc.rs:227-237` (comment: "does NOT restore the data-plane state machine … silent data-loss-shaped bug")

## Failure scenario
1. A follower falls behind the leader's log-retention/compaction window.
2. Leader sends `InstallSnapshot`; follower's Raft `last_applied` jumps to N.
3. Follower's data plane still reflects only pre-snapshot state → **permanent silent divergence** between Raft index and actual data. Queries wake and read missing state.

## Fix direction
Either (a) transfer + restore real state-machine bytes for data groups, or (b) refuse to advance `last_applied` when no data payload is present (fail loud instead of diverging). Add a test that compacts past a follower and verifies the follower's *data* matches the leader.

*v1.0 production-hardening, phase P2.*

farhan-syah:
Fixed in bff3384b — though the investigation found the root cause differs from the mechanism described in the report, so this is worth spelling out.

**What the report described:** data-group `InstallSnapshot` advancing the Raft index without transferring state-machine data.

**What the investigation actually found** (via an end-to-end learner catch-up: create a collection, write rows, join a fresh learner):

1. The **data**-group snapshot path does transfer state — `build_group_snapshot` ships the documents and the PK→surrogate map, and `apply_snapshot` reinstalls them on the learner. The learner's data was present; that half was not the bug.
[...]

PR #247, opened 2026-08-23
fix(raft): P2 consensus safety — check-quorum, leader lease, seed commit_index, learner gate

### LINKED ISSUE #161: Raft HardState (current_term/voted_for) is never persisted — split-brain on restart
**Source:** static engine audit, 2026-07-05 (cluster/consensus lens), verified against `main @ d92ede42`
**Severity:** Critical — split-brain / committed-entry divergence after a restart.

## Summary
Raft's `HardState` (`current_term`, `voted_for`) is **never durably persisted**. `persist_hard_state` only *stages* it into `Ready.hard_state`, and no consumer ever writes it: `save_hard_state` has **zero production callers** (verified by full-repo grep — only trait def, impls, and `#[cfg(test)]` roundtrips). The multi-raft tick loop reads only `.messages`/`.committed_entries`/`.snapshots_needed` from `Ready` and drops `.hard_state`.

Compounding it: even the staged state would be flushed on the *next tick*, but vote/append RPC responses are returned on the wire **before** that — violating Raft's "persist to stable storage before replying" rule.

## Location
- `nodedb-raft/src/node/internal.rs:366` (stages only)
- `nodedb-cluster/src/raft_loop/tick.rs:55-75` (consumer ignores `.hard_state`)
- `nodedb-cluster/src/raft_loop/handle_rpc.rs:70-79` (replies before persist)
- `nodedb-raft/src/storage.rs:28` (`save_hard_state` — no prod caller)

## Failure scenario
1. Node votes for candidate A in term 5 → `voted_for=A, current_term=5` held **in memory only**.
2. Node crashes and restarts → `load_hard_state` returns `HardState::default()` (term 0, voted_for 0).
3. Candidate B requests a vote in term 5 → node has no record of having voted → grants it.
4. **Two leaders in term 5** → divergent committed entries. Term also regresses to 0 on every restart (monotonicity violation), only re-learned from an incoming leader.

## Fix direction
Persist `HardState` to stable storage **in the RPC path, before** returning any vote/append response (drain `Ready.hard_state` → `save_hard_state` synchronously). Add a test: "vote A, crash, restart, must NOT grant a vote to B in the same term."

*v1.0 production-hardening, phase P2 (cluster consensus safety).*

EnRaiha:
P2 core consensus-safety fixes for epic #165 (Critical #161/#162 proofs, High #3 deposed-leader reads, High #4 bounded staleness). 9 fix commits + 4 docs commits.

- check-quorum: leader steps down after election_timeout_max without quorum contact (052f2fa64).
- leader lease for linearizable Strong reads, starts at election win (cf2cd5c24 + 37ea1898f).
- seed commit_index from durable floor on restore — GAP-4 fix (dca42ac89).
- enforce learner catch-up before promotion, propose-side (b215d4bba).
- regression + proof tests for #161 (persist-before-reply) and #162 (apply-before-advance) (da49bf0f2).
- Docs: fsync audit, consensus reviews, GAP-4 double-review verdict, full plan + post-P2 roadmap + phase-2 report (P2/).

Heavy test: raft 105/105 + integration 9+7+8, cluster all-features 1015, clippy 0 (7 consecutive green runs).

Closes #161, #162. Part of #165.


farhan-syah:
Thanks for this. Closing the branch, but the work isn't lost — I took it into `main` piece by piece instead of merging.

**Landed** in 7480c1b86, with you as co-author:

- seed `commit_index` from the durable applied floor on restart
- the restart double-vote and snapshot apply-ordering regression tests

**Landing separately:** the learner catch-up gate. Propose-side placement is right, and your reasoning about apply-time `commit_index` having already passed the promote entry is the part I'd have got wrong.

**Reworked rather than taken:** check-quorum. The step-down is correct, but the contact refresh reads `match_index >= last_log_index - 1` from inside `try_advance_commit_index`, which is only reached on a successful voter response. Two healthy cases lose contact there:

- a follower that is reachable but rejecting, backtracking through a log conflict
- a follower that is merely lagging, when a write burst outruns the acks in flight

Either way a leader with an intact quorum gets deposed. `LeaderState::record_ack` is the signal that answers the actual question — it's bumped on success and rejection alike, and `read_index` already uses it for exactly this.

**Not taken:** the leader lease. No read path calls it, and `read_index_wait::confirm_read_index` already serves linearizable reads without assuming bounded clock drift. A lease trades that safety property for latency, so I'd rather not make Strong reads clock-dependent.

The 17 markdown files can't land — planning notes stay out of the tree.

PR #249, opened 2026-08-25
refactor(lease): inject WallClock trait into count_matching_leases

EnRaiha:
## Summary
Lease expiry is a *duration* question, so it must be measured against real wall time. PR #246 established this after the `HlcClock::peek()` wedge (a frozen HLC on an idle cluster makes every lease look unexpired). This extracts the clock behind a `WallClock` trait so the expiry check in `count_matching_leases` is injectable.

- New `nodedb/src/control/lease/clock.rs`: `WallClock` trait, `RealWallClock` (production), `MockClock` (tests).
- `count_matching_leases` now takes `&dyn WallClock` instead of calling `super::wall_now_ns()` directly.
- `wait_for_lease_drain` passes `&RealWallClock`.
- Adds 3 `MockClock` tests that pin the clock and prove expiry uses the injected wall clock, not the HLC (including the skewed-HLC case that must not drop a live lease).

## Scope
Single logical change: clock injection for lease expiry. No behaviour change in production (real wall time is still used). Under ~80 lines of new code + tests.

## Test plan
`cargo test -p nodedb --lib control::lease::drain_propose` (existing + 3 new tests).

Follow-up to #246.

### DIFF
--- a/nodedb/src/control/lease/clock.rs
+++ b/nodedb/src/control/lease/clock.rs
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+//! Injectable wall-clock source for lease expiry checks.
+//!
+//! Lease expiry is a *duration* question — "have N nanoseconds passed?" — so it
+//! must be measured against real wall time, never against a Hybrid Logical
+//! Clock. An HLC only advances on a local event or an inbound message, so on an
+//! idle cluster it physically freezes; comparing a lease's `expires_at` against
+//! `HlcClock::peek()` would find every lease unexpired and reinstate the wedge
+//! (see PR #246 / `drain_propose.rs`).
+//!
+//! The [`WallClock`] trait lets production code read the real clock while tests
+//! drive expiry deterministically with a [`MockClock`] instead of mocking
+//! `SystemTime` or depending on `HlcClock::peek`.
+
+use std::sync::atomic::{AtomicU64, Ordering};
+
+/// A source of nanoseconds since the Unix epoch.
+pub trait WallClock {
+    fn now_ns(&self) -> u64;
+}
+
+/// Reads the real system clock.
+pub struct RealWallClock;
+
+impl WallClock for RealWallClock {
+    fn now_ns(&self) -> u64 {
+        super::wall_now_ns()
+    }
+}
+
+/// Deterministic clock for tests. Set it explicitly so expiry assertions do not
+/// depend on `SystemTime` or on `HlcClock::peek` (which never advances on its
+/// own and would make every lease look unexpired).
+#[cfg(any(test, feature = "test-helpers"))]
+pub struct MockClock {
+    now: AtomicU64,
+}
+
+#[cfg(any(test, feature = "test-helpers"))]
+impl MockClock {
+    pub fn new(ns: u64) -> Self {
+        Self {
+            now: AtomicU64::new(ns),
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


farhan-syah:
Closing: the wedge is already fixed on `main`, and the tests here duplicate tests that already exist.

**Correction to my earlier comment on this PR:** it credited the fix to #246. That is wrong. #246 was closed unmerged on 2026-08-24. The fix reached `main` as a separate commit, cited below. The rest of that comment stands, and is restated here in full.

## The fix is on `main` as 7168ecc55

`7168ecc55` — *fix(control): drop lease-drain wedges from stale or expired holders*, 2026-08-24 — added the membership filter and the wall-time expiry comparison directly. `count_matching_leases` already reads wall time, not the HLC:

- `nodedb/src/control/lease/drain_propose.rs:256` — `let now_wall_ns = super::wall_now_ns();`
- Rationale documented at `drain_propose.rs:235-243`.

That commit predates this PR by a day. This PR changes no production behaviour, as its own Scope section states. What remains is a test seam.

## The three new tests duplicate existing ones

`7168ecc55` also added four tests, three of which cover the same ground:

| This PR | Already on `main` |
| --- | --- |
| `wall_clock_expired_lease_is_dropped` | `expired_lease_does_not_block_drain_count` (`drain_propose.rs:493`) |
| `wall_clock_ignores_skewed_hlc` | `a_live_lease_still_blocks_when_the_hlc_runs_ahead_of_wall_time` (`drain_propose.rs:553`) |
| `wall_clock_live_lease_is_counted` | `member_unexpired_lease_still_blocks_drain_count` (`drain_propose.rs:583`) |

The fourth, `expired_lease_stops_blocking_even_with_an_unadvanced_hlc` (`drain_propose.rs:513`), has no counterpart in this PR.

The existing fixtures derive expiry from real `wall_now_ns()` with a 60-second margin. That is deliberate, and the comment at `drain_propose.rs:454` says why: a fixture built from the same clock the code reads puts both sides in one frame and the assertion holds whatever the comparison does. `MockClock` reintroduces that. A test that pins the clock the code reads proves the code reads the injected value, not that it reads wall time.

## Design problems in the seam itself

- **Inverted layering.** `RealWallClock::now_ns` in `nodedb/src/util/wall_clock.rs` calls `crate::control::lease::wall_now_ns()`, so `util` depends upward on `control`. The PR widens `pub(super) use wall_time::wall_now_ns` to `pub(crate)` to allow it. If the abstraction belongs in `util`, `wall_now_ns` moves there and `control::lease` calls down.
- **One implementation, one call site.** `wait_for_lease_drain` hardcodes `&RealWallClock`. Nothing else injects. This adds `&dyn` dispatch to a poll loop for tests that already pass.
- **Dead API.** `MockClock::set` is never called.
- **Churn.** Eight call sites rewritten to pass `&clock_now()`, a helper that reads real wall time and hands it straight back.

## It no longer applies

`f19756c96` (2026-08-28) added the `own_holds: u32` self-exclusion parameter to `count_matching_leases` (`drain_propose.rs:250-282`). The signature has moved and this branch conflicts.

## The correct fix

A crate-wide clock source in `util` or `nodedb-types`, owning `wall_now_ns` outright, with `renewal::tick` (`nodedb/src/control/lease/renewal.rs:176`) calling through it as well. That gives one source, no upward dependency, and a second caller that earns the indirection — and it would let renewal tests advance time instead of sleeping. That is a separate PR, and no current test is blocked on it.

PR #251, opened 2026-08-25
feat(lease): release leases on SWIM-Dead via lease GC hook

EnRaiha:
## Summary
Closes epic #165 item 2. A crashed node never produces `TopologyChange::Leave`, so its leases survived until the metadata leader's periodic sweep (~60s, leader-only). `LeaseGcOnCrashHook` subscribes to SWIM membership and spawns a grace-windowed lease release the instant a member is confirmed `Dead`.

**Stacked on #249 + #250** (WallClock trait → MAX_SKEW clamp): this branch contains both first; the diff reduces once they merge.

### Changes
- `gc_leases_for_node_if` — predicate-filtered release (review Option 1); `gc_leases_for_node` forwards with `|_| true`. Reset `nodedb/src/control/lease/gc.rs`.
- `LeaseGcOnCrashHook` — `Weak<SharedState>`, `is_singleton_worker()` guard, `tokio::spawn` (never inline propose-and-wait; the #246 deadlock lesson), `debug!` on non-numeric `NodeId` skip (`seed:…` placeholders).
- **Grace window** (`MAX_SKEW_NS`, 5 min): only leases expiring within the window are released on Dead; fresh leases are spared for the periodic sweep — a false-positive Dead (live-but-partitioned) must not yank a live hold (the #246 correctness line). Leases beyond the window are only swept once the node remains a non-member.
- **Plumbing**: `nodedb_cluster::start_cluster_subsystems` accepts `extra_swim_subscribers`, threaded into `SwimSubsystem::new` at `bootstrap/start.rs`; the host builds the hook in `start_raft/loop_build.rs`.
- `MAX_SKEW_NS` → `pub(crate)` in `drain_propose.rs`.

### Semantics (matching review's risk acceptance)
Leases expiring more than MAX_SKEW in the future are NOT released on SWIM-Dead; the periodic sweep handles them only if the node remains Dead (non-member).

### Verification (solve-first)
- `cargo check -p nodedb-cluster --tests` — EXIT 0
- `cargo check -p nodedb --tests` — EXIT 0
- `cargo test -p nodedb --lib lease::gc` — 8/8 PASSED (4 existing + 4 new: parse, filter, Dead-only decision, grace-window selection via `RecordingProposer`)
- `cargo test -p nodedb --lib lease::drain_propose` — 15/15 PASSED

Note: cluster-tests integration (kill node → leases released < sweep interval) is not in this PR; it needs the multi-node harness and will follow as its own test PR.

### DIFF
--- a/nodedb-cluster/src/bootstrap/start.rs
+++ b/nodedb-cluster/src/bootstrap/start.rs
@@ -34,6 +34,7 @@ use crate::subsystem::{
     DecommissionSubsystem, ReachabilitySubsystem, RebalancerSubsystem, RunningCluster,
     SubsystemRegistry, SwimSubsystem, SwimSubsystemConfig,
 };
+use crate::MembershipSubscriber;
 use crate::swim::config::SwimConfig;
 use crate::swim::incarnation::Incarnation;
 use crate::swim::incarnation_store::CatalogIncarnationStore;
@@ -73,6 +74,7 @@ pub fn register_default_subsystems(
     ctx: &BootstrapCtx,
     executor: Arc<MigrationExecutor>,
     catalog: &Arc<ClusterCatalog>,
+    extra_swim_subscribers: Vec<Arc<dyn MembershipSubscriber>>,
 ) -> crate::error::Result<()> {
     // Fast-restart rejoin safety: resume SWIM above the last persisted
     // incarnation. A node that crashed while peers held `Dead(A, N)`
@@ -112,7 +114,7 @@ pub fn register_default_subsystems(
         swim_cfg,
         Arc::clone(&ctx.routing),
         Arc::clone(&ctx.topology),
-        vec![],
+        extra_swim_subscribers,
     )));
 
     registry.register(Arc::new(ReachabilitySubsystem::new(
@@ -208,6 +210,7 @@ pub async fn start_cluster_subsystems(
     transport: Arc<NexarTransport>,
     raft_multi_raft: Arc<std::sync::Mutex<crate::multi_raft::MultiRaft>>,
     catalog: &Arc<ClusterCatalog>,
+    extra_swim_subscribers: Vec<Arc<dyn MembershipSubscriber>>,
 ) -> Result<RunningCluster> {
     let health = ClusterHealth::new();
     let (decommission_signal, _) = tokio::sync::watch::channel(false);
@@ -228,7 +231,7 @@ pub async fn start_cluster_subsystems(
     ));
 
     let mut registry = SubsystemRegistry::new();
-    register_default_subsystems(&mut registry, config, &ctx, executor, catalog)?;
+    register_default_subsystems(&mut registry, config, &ctx, executor, catalog, extra_swim_subscribers)?;
 
     registry
         .start_all(&ctx)
--- a/nodedb/src/control/cluster/start_raft/loop_build.rs
+++ b/nodedb/src/control/cluster/start_raft/loop_build.rs
@@ -165,6 +165,14 @@ pub(super) fn build_raft_loop(
         scheduler_config: &scheduler_config,
     })?;
 
+    // SWIM-Dead → lease GC hook: release a crashed node's leases the moment
+    // SWIM confirms it Dead (grace-windowed), instead of waiting for the
+    // metadata leader's periodic sweep.
+    let extra_swim_subscribers: Vec<std::sync::Arc<dyn nodedb_cluster::MembershipSubscriber>> =
+        vec![std::sync::Arc::new(
+            crate::control::lease::gc::LeaseGcOnCrashHook::new(&shared),
+        )];
+
     let running = tokio::task::block_in_place(|| {
         tokio::runtime::Handle::current().block_on(nodedb_cluster::start_cluster_subsystems(
             &pending.config,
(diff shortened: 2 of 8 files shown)


farhan-syah:
Closing. The hook is the right piece and it is built correctly — but it is wired to undo a clamp that is itself closed, and the grace window it is built around releases every lease anyway.

## What this PR gets right

`LeaseGcOnCrashHook` mirrors the shipped `TopologyChange::Leave` hook exactly (`metadata_applier/dispatch.rs:274-290`): `Weak<SharedState>`, `tokio::spawn` rather than an inline propose-and-wait, `is_singleton_worker()` guard. The subscriber contract sanctions it — "Heavier work belongs on a dedicated task the subscriber spawns itself" (`swim/subscriber.rs:11-13`) — and the detector loop is async (`detector/runner.rs:219`), so the spawn has a runtime. `gc_leases_for_node_if` is a clean extraction. Nothing below is a criticism of that work.

## Root cause

`expires_at` is stamped in the holder's clock frame and compared in the checking node's, and nothing merges those two frames.

- Stamped: `compute_expires_at(hlc_clock.now(), duration)`, `lease/propose.rs:194`.
- Compared: `l.expires_at.wall_ns > super::wall_now_ns()`, `lease/drain_propose.rs:256`.
- Never merged: `HlcClock::update` has one production caller, `catalog_entry/descriptor_stamp.rs:65`, which folds only the prior catalog entry's `modification_hlc` for the same descriptor. No Raft, cluster, or lease path feeds it. Filed as #264.

Two independent monotone clocks, compared as if they shared an origin. No constant makes that comparison correct, because the error term is another machine's offset and is unbounded by construction.

Every PR in this stack is a different attempt to make that comparison trustworthy:

| PR | Attempt | Outcome |
| --- | --- | --- |
| #249 | Inject the clock behind a trait | The comparison is unchanged; only the test seam moves |
| #250 | Widen the comparison by `MAX_SKEW` | Picks 300 s against a 35 s `DEFAULT_DRAIN_TIMEOUT`, so a crashed holder fails DDL for five minutes |
| #251 | Add a hook to release early and undo the widening | The window admits every lease, so it undoes nothing |

Three layers, each correcting the previous one, none touching the frame mismatch underneath.

## Why the window is a no-op

| Constant | Value | Source |
| --- | --- | --- |
| `MAX_SKEW_NS` | 300 s | `drain_propose.rs:23` |
| `DEFAULT_LEASE_DURATION` | 300 s | `lease/propose.rs:16` |
| `descriptor_lease_duration_secs` default | 300 s | `nodedb-types/src/config/tuning/network.rs:280` |

A lease is stamped `holder_now + 300s`, and the filter accepts `expires_at <= now + 300s`. Every lease satisfies that at every moment of its life, unless the holder's clock runs ahead of the checker's.

So the stated property — "a false-positive Dead must not yank a fresh lease" — is not delivered. The only leases spared are the ones from clock-ahead peers, spared by exactly that peer's offset. The window's single observable effect is governed by the same cross-node skew the stack exists to defend against.

## The fallback it defers to does not exist

`collect_non_member_leases` (`gc.rs:25`) has no production caller; the only references are its own unit tests. The one match for "periodic sweep" in the tree is a log string at `dispatch.rs:287`.

The window's rationale is that "leases beyond the window are left to the periodic sweep". Nothing collects them. `main`'s module doc already made this claim, so the PR inherited it rather than introducing it — but this design is the first thing to depend on it.

## One point in the PR's favour, and where it actually belongs

A false-positive Dead does not self-heal. `collect_near_expiry` (`renewal.rs:177`) re-acquires only leases still present in the cache, so once an entry is released it is gone and the renewal loop never sees it.

The backstop for that is descriptor-version fencing at COMMIT — `gateway::version_check::check_descriptor_holds`, per #165 — not a window in the GC. That is why the Leave hook already releases unconditionally. A partitioned node wrongly declared Dead loses its lease and its stale plan is caught at the fence, which is the designed behaviour.

## The end state

In order. Each step removes the reason the next one was ever needed.

1. **Release on SWIM-Dead, unfiltered.** This PR's hook calling `gc_leases_for_node`. Crash recovery becomes an authoritative membership event instead of a timestamp read. Delete `gc_leases_for_node_if`, `near_expiry_filter`, and the `MAX_SKEW_NS` dependency.
2. **Add the sweep the doc has always claimed**, over non-member holders, or delete the claim. Right now it is neither.
3. **Wire `HlcClock::update` at every cluster ingress (#264).** Only then do `expires_at` and `now` share a frame, and only then is a skew bound derivable from measured offset rather than invented.

After 1 and 2, expiry is what `count_matching_leases` already documents it as: a backstop for a live-but-wedged holder, not the crash-recovery path. After 3 it is a comparison that means something. No `MAX_SKEW` constant and no `WallClock` trait appear anywhere in that sequence.

## Also

- `mergeable: CONFLICTING`, and 227 of the 518 added lines belong to #249 and #250.
- `start_cluster_subsystems` and `register_default_subsystems` are `pub` and re-exported at `nodedb-cluster/src/lib.rs:81`. Adding a parameter breaks a published crate's API — acceptable at 0.5.0, needs the version bump.
- The lease filter moved from the cache key's holder to `l.node_id`. They are equal at every grant (`propose.rs:193` and `:198`), so this is safe today, and it silently retargets the GC if they ever diverge.

Step 1 is the remaining #165 item.

PR #252, opened 2026-08-25
feat(trigger): run Event-Plane trigger bodies as one system transaction

EnRaiha:
## Summary
Related to #165 (deferred actions — last unchecked box). A trigger body that fails mid-block leaves earlier statements applied, and its async retry repeats them. The Event-Plane fire path now stages every fired body into a `SystemTxnScope`: all statements commit together or roll back together.

**Option B′ (scope-bound executor)**, per the Phase-1 recon: bodies are procedural control flow (`If`/`While`/`For`), so pre-planning the whole body (Option A) is impossible; instead each statement's tasks route into the scope as they are planned.

## Gap hardening (review items 1–3)
1. **Lease-scope lifetime** — `SystemTxnScope::retain_lease_scope` holds each statement's `Arc<QueryLeaseScope>` until COMMIT's version fence.
2. **Cascade sharing** — `system_scope` threaded through `FireTriggersParams` + all `FireAfter*Params`; `StatementExecutor::with_system_scope`; nested (cascade) triggers inherit the caller's scope — one failure rolls back every body the cascade touched.
3. **DDL fail-fast** — `is_system_ddl` rejects the one DDL-shaped Data-Plane op (`MetaOp::ConvertCollection`) with `SystemTxnError::Ddl` before staging.

## Mechanics
- `push_task_into_scope` (statement-serial staging, extracted from `run_tasks_atomically`; the batch function is refactored onto it — behavior identical)
- `commit_scope` / `rollback_scope` exported helpers
- `dispatcher/single.rs`: one scope per fired event (**async only** — `EventSource::User`); row + statement fires share it; commit on `FireReport::has_failure() == false`, rollback otherwise
- Sync / DEFERRED / BEFORE / INSTEAD paths explicitly `system_scope: None` — never double-wrapped (they already run inside the client transaction)
- New non-consuming `FireReport::has_failure()`

## Verification (solve-first)
- `cargo check -p nodedb --lib --tests` — EXIT 0, zero warnings
- `control::system_txn` — 1/1 (DDL gate)
- `control::trigger` — 68/68 (no regression from the threading)
- `event::trigger` — 12/12 (dispatcher wiring)

Known follow-ups (tracked, not blockers): retry-path atomicity wiring in `retry_action.rs` (currently `None` — retries run the failed body non-atomically until it lands), cluster-tests stage-2 coverage (kill mid-trigger → retry/DLQ), deferred `batch.rs` path.

### DIFF
--- a/nodedb/src/control/planner/procedural/executor/core/dispatch.rs
+++ b/nodedb/src/control/planner/procedural/executor/core/dispatch.rs
@@ -102,7 +102,24 @@ impl<'a> StatementExecutor<'a> {
             })
             .await?;
 
-        if let Some(ref tx_ctx) = self.tx_ctx {
+        if let Some(ref system_scope) = self.system_scope {
+            // System-transaction mode (Event-Plane trigger fire): stage every
+            // task into the active scope instead of dispatching immediately.
+            // All statements of the fired body commit or roll back together;
+            // the scope retains each plan's lease scope until COMMIT.
+            let lease_scope = std::sync::Arc::new(lease_scope);
+            for task in tasks {
+                crate::control::system_txn::push_task_into_scope(
+                    system_scope,
+                    self.state,
+                    task,
+                    std::sync::Arc::clone(&lease_scope),
+                    self.event_source,
+                )
+                .await
+                .map_err(|e| crate::Error::BadRequest { detail: e.to_string() })?;
+            }
+        } else if let Some(ref tx_ctx) = self.tx_ctx {
             let mut guard = tx_ctx.lock().unwrap_or_else(|p| p.into_inner());
             guard.buffer_statement(tasks, lease_scope);
         } else {
--- a/nodedb/src/control/planner/procedural/executor/core/mod.rs
+++ b/nodedb/src/control/planner/procedural/executor/core/mod.rs
@@ -59,6 +59,12 @@ pub struct StatementExecutor<'a> {
     /// Arc<Mutex> required (not RefCell) because execute_statement returns `+ Send` futures.
     pub(super) new_mutations: Arc<Mutex<HashMap<String, nodedb_types::Value>>>,
     pub(super) tx_ctx: Option<Arc<Mutex<ProcedureTransactionCtx>>>,
+    /// System transaction scope for the Event-Plane trigger fire path.
+    /// When set, `execute_sql` stages every task into this scope
+    /// (all-or-nothing per fired body) instead of dispatching immediately.
+    /// Nested (cascade) executors inherit the same scope so one failed body
+    /// rolls back every body the cascade touched.
+    pub(super) system_scope: Option<Arc<crate::control::system_txn::SystemTxnScope>>,
     pub(super) out_values: Arc<Mutex<HashMap<String, nodedb_types::Value>>>,
     /// Cross-shard origin context; `Some` only in the Event-Plane trigger fire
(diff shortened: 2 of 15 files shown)


farhan-syah:
Closing. The scope is one transaction per event, but the retry queue is keyed per trigger. That mismatch discards successful triggers' writes.

## The loss path

Two triggers fire on one event under `FireErrorPolicy::Continue`, sharing the scope.

1. Trigger A stages its writes and succeeds.
2. Trigger B fails.
3. `has_failure()` is true, so `rollback_scope` discards the scope. A's writes are gone.
4. `record_row_failures` enqueues failures only — `into_failures()` filters on `outcome.error.is_some()` (`enqueue.rs:64`, `fire_common.rs:111`). A is not enqueued.
5. The retry re-fires B alone (`only_trigger`).

A never runs again, and nothing records that it was undone. Today A's writes land and stay.

## The invariant is documented at the line being broken

`fire_common.rs:65-69`, on the `Continue` variant this path passes:

> Run every trigger and report each outcome. The Event Plane uses this: its triggers share no transaction, so one failing must not silently cancel its siblings — they would never run at all, since a retry re-fires only the trigger that failed.

The PR makes them share a transaction. It leaves `Continue` and the per-trigger retry key untouched.

## The shipped pattern is one scope per retry unit

`event_trigger.rs:361` wraps one action, and the action is the retry unit (#165, "Retry unit is one action, not one source write"). Its comment states the rule: "re-running a half-applied action repeats the tasks that already landed — which is what makes a retry queue unsafe to point at it."

Scope granularity must equal retry granularity.

## Other defects

- **DEFERRED excluded on a false premise.** The guard reads `event.source == EventSource::User`, commented "Sync and DEFERRED fires must NOT be wrapped — they already run inside the client's transaction". `event/types.rs:162` defines `Deferred` as post-commit from a transaction batch. That transaction has committed. Sync, BEFORE and INSTEAD never reach `dispatch_triggers` at all — they go through `dml_hook_fire`.
- **The item is not closed.** `retry_action.rs` passes `system_scope: None` at both sites, so a retried body still half-applies and is retried again. That is the second half of #165's "its retry repeats what landed". `batch.rs` is `None` too.
- **Failures degrade to the bug.** `SystemTxnScope::begin` failing warns and continues "firing without atomicity". `commit_scope` failing warns only — the event counts as delivered, nothing committed, nothing enqueued.
- `cargo fmt` was not run. Three `system_scope,` lines in `single.rs` are indented to 8 columns inside blocks at 20.

## The correct fix

One scope per trigger body, not per event.

That matches the shipped event-action pattern, keeps scope granularity equal to retry granularity, preserves sibling independence under `Continue`, and makes each body atomic — what #165 asks for, and nothing more.

Most of this PR survives. `push_task_into_scope`, `retain_lease_scope`, the `is_system_ddl` gate, the `with_system_scope` threading for nested cascade triggers, and `has_failure()` all stay as written. What moves is where `begin`, `commit` and `rollback` sit: into the per-trigger loop in `fire_common`, not around the whole dispatch in `single.rs`.

The `EventSource::User` guard then disappears. A body is a body whatever sourced the event, so DEFERRED, retry and batch get atomicity instead of each needing its own `system_scope: None`.

Two changes go with it:

- `begin` and `commit_scope` failures enqueue the trigger for retry. A body that could not be made atomic has not run.
- The retry path passes a scope like every other caller. That closes the second half of the #165 item.

PR #259, opened 2026-08-26
fix(pgwire)

### LINKED ISSUE #254: multi-statement SQL batches silently drop statements beyond 32 (no error)
## Bug Description

multi-statement SQL batches silently drop statements beyond 32 (no error)

## Steps to Reproduce

1. Send 64 UPSERT statements in one multi-statement query via pgwire.
2. `SELECT count(*)` → 32 (half landed). 128 statements → ~25% landed.
3. No error returned — client believes the batch committed.

## Expected Behavior

All statements commit, or the client receives a hard error on partial commit.

## Actual Behavior

Silent divergence between what the client thinks committed and what is durable. Measured on 2026-08-26 during code-graph import: 64→50% land, 128→25%, 32→100%.

## Proposed Fix

`sql_exec.rs` multi-statement branch: after executing all statements, if `all.len() < total` → `XX000 "incomplete batch: N/M statements produced responses — some work may have been dropped; retry with a smaller batch"`. Defense-in-depth until the propose-path drop site (vshard channel full backpressure) is traced and fixed.

Files: sql_exec.rs.

EnRaiha:
## Bug Description

multi-statement SQL batches silently drop statements beyond 32 (no error)

## Steps to Reproduce

1. Send 64 UPSERT statements in one multi-statement query via pgwire.
2. `SELECT count(*)` → 32 (half landed). 128 statements → ~25% landed.
3. No error returned — client believes the batch committed.

## Expected Behavior

All statements commit, or the client receives a hard error on partial commit.

## Actual Behavior

Silent divergence between what the client thinks committed and what is durable. Measured on 2026-08-26 during code-graph import: 64→50% land, 128→25%, 32→100%.

## Proposed Fix

`sql_exec.rs` multi-statement branch: after executing all statements, if `all.len() < total` → `XX000 "incomplete batch: N/M statements produced responses — some work may have been dropped; retry with a smaller batch"`. Defense-in-depth until the propose-path drop site (vshard channel full backpressure) is traced and fixed.

Files: sql_exec.rs.

Closes #254

### DIFF
--- a/nodedb/src/control/server/pgwire/handler/sql_exec.rs
+++ b/nodedb/src/control/server/pgwire/handler/sql_exec.rs
@@ -43,10 +43,30 @@ impl NodeDbPgHandler {
                     .await
             }
             _ => {
+                let total = statements.len();
                 let mut all = Vec::new();
                 for stmt in statements {
                     let mut resp = self.execute_single_sql(identity, session_id, &stmt).await?;
                     all.append(&mut resp);
+                }
+                // B1 ack reconciliation (2026-08-26): every statement must
+                // produce at least one response. A short count means the
+                // executor silently dropped work (observed: batch 64 → 32
+                // landed, batch 128 → ~25%, no error). Surface it as a hard
+                // error instead of letting the client believe the batch
+                // committed. Defense-in-depth until the propose-path drop
+                // site is traced and fixed.
+                if all.len() < total {
+                    return Err(PgWireError::UserError(Box::new(ErrorInfo::new(
+                        "ERROR".to_owned(),
+                        "XX000".to_owned(),
+                        format!(
+                            "incomplete batch: {}/{} statements produced responses — \
+                             some work may have been dropped; retry with a smaller batch",
+                            all.len(),
+                            total,
+                        ),
+                    ))));
                 }
                 Ok(all)
             }


farhan-syah:
Closing. The check cannot fire for the bug it targets.

## Why it is inert

```rust
for stmt in statements {
    let mut resp = self.execute_single_sql(identity, session_id, &stmt).await?;
    all.append(&mut resp);
}
```

Each statement is awaited, and errors propagate with `?`. #254 reports no error and a client that believes the batch committed, so `execute_single_sql` returned `Ok` with responses for all 64 while only 32 rows became durable.

`all.len() == total`. The guard passes. An acknowledged-but-not-durable write produces responses like any success.

For `all.len() < total` to hold, a statement must return zero responses. That is a different failure from the one filed.

## The comparison is also wrong in units

`all` counts responses; `total` counts statements. One statement returning two responses masks another returning none.

## The suspected cause is ruled out

The PR names vshard channel full backpressure. That path does drop on a full bounded channel (`calvin/scheduler/driver/core/catch_up.rs:6-11`), but `channel_capacity` defaults to 512, not 32 (`calvin/scheduler/driver/config.rs:38`), and it already recovers through `drain_catch_up`.

## 32 is exact, so it is not backpressure

64 → 32, 128 → 32, 32 → 32. A full queue drops a count that varies with timing and load. A constant 32 is a fixed-size bound. `split_sql_statements` has no cap, so the truncation sits downstream of the split and upstream of durability.

That exactness is the strongest evidence available for locating the drop site. This PR spends it on a guard instead of following it.

## The correct fix

Find the drop site first. Acknowledgement is decoupled from durability somewhere between `execute_single_sql` returning `Ok` and the write landing, and the fixed 32 says where to look.

A guard afterwards must compare durable outcomes to statements, not response-vector length, and it belongs where the acknowledgement is issued rather than in the pgwire batch loop.

The title is `fix(pgwire)` with no subject. The error text advises "retry with a smaller batch" — a client workaround for server-side data loss, on a branch the reported case never reaches.

#254 stays open.

PR #260, opened 2026-08-26
feat(health)

### LINKED ISSUE #255: data-plane core wedge produces zero ERROR log lines and /healthz stays green
## Bug Description

data-plane core wedge produces zero ERROR log lines and /healthz stays green

## Steps to Reproduce

1. Create sustained load that jams a data-plane core (e.g. FTS index rebuild during writes).
2. Core stops ticking; requests hang.
3. Journal shows no ERROR (only `checkpoint response timed out core_id=0`); /healthz keeps reporting ok; every query hangs on descriptor-lease timeout.

## Expected Behavior

Health endpoint reflects runtime stall so operators get alerted.

## Actual Behavior

Full wedge with 0 ERROR lines, node unreachable, health endpoint still green. Observed 2026-08-26 12:23–12:38: full wedge, node unreachable. Only trace was a single checkpoint timeout.

## Proposed Fix

New `bridge/runtime_health.rs`: `record_core_tick` on every event-loop iteration (`event_loop.rs`), `record_checkpoint_result` (3× consecutive failure → stall flag). `HealthState::Degraded { reason }` → HTTP 503 `{"status":"degraded"}` + native STATUS "Degraded". Tests included.

Files: runtime_health.rs (new), bridge/mod.rs, event_loop.rs, health.rs.

EnRaiha:
## Bug Description

data-plane core wedge produces zero ERROR log lines and /healthz stays green

## Steps to Reproduce

1. Create sustained load that jams a data-plane core (e.g. FTS index rebuild during writes).
2. Core stops ticking; requests hang.
3. Journal shows no ERROR (only `checkpoint response timed out core_id=0`); /healthz keeps reporting ok; every query hangs on descriptor-lease timeout.

## Expected Behavior

Health endpoint reflects runtime stall so operators get alerted.

## Actual Behavior

Full wedge with 0 ERROR lines, node unreachable, health endpoint still green. Observed 2026-08-26 12:23–12:38: full wedge, node unreachable. Only trace was a single checkpoint timeout.

## Proposed Fix

New `bridge/runtime_health.rs`: `record_core_tick` on every event-loop iteration (`event_loop.rs`), `record_checkpoint_result` (3× consecutive failure → stall flag). `HealthState::Degraded { reason }` → HTTP 503 `{"status":"degraded"}` + native STATUS "Degraded". Tests included.

Files: runtime_health.rs (new), bridge/mod.rs, event_loop.rs, health.rs.

Closes #255

### DIFF
--- a/nodedb/src/bridge/mod.rs
+++ b/nodedb/src/bridge/mod.rs
@@ -4,6 +4,7 @@ pub mod admission_chokepoint;
 pub mod dispatch;
 pub mod envelope;
 pub mod quiesce;
+pub mod runtime_health;
 pub mod slab;
 
 // Re-export shared query engine from nodedb-query crate.
--- a/nodedb/src/bridge/runtime_health.rs
+++ b/nodedb/src/bridge/runtime_health.rs
@@ -1,0 +1,144 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+//! Runtime health registry — per-core heartbeat ticks + checkpoint stall flag.
+//!
+//! B4/B5 fix (2026-08-26): the 12:23 wedge showed zero ERROR logs while the
+//! data-plane core loop and the coordinated checkpoint silently stalled. The
+//! startup gate kept reporting `Ok` because it only tracks boot phases, so
+//! `/healthz` and `STATUS` had no idea the node was wedged.
+//!
+//! Every event-loop iteration records a monotonic tick per core; the health
+//! formatter compares ticks against a stall threshold. The checkpoint manager
+//! raises a stall flag after consecutive failed cycles. Both signals are
+//! surfaced through [`crate::control::startup::health::HealthState::Degraded`]
+//! so HTTP `/healthz` degrades to 503 and the native `STATUS` reports
+//! "Degraded" — a wedge becomes loud instead of silent.
+//!
+//! Lives in `bridge` so both the data plane (writer) and control plane
+//! (reader) can reach it without a control→data or data→control dependency.
+
+use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
+use std::sync::{LazyLock, Mutex};
+use std::time::{Duration, Instant};
+
+/// A core whose last recorded tick is older than this is considered stalled.
+///
+/// The event loop polls at most every 100 ms (`IDLE_POLL_TIMEOUT_MS`), so a
+/// healthy core ticks at least 10×/s. 60 s of silence means the loop is stuck
+/// inside a request or maintenance work — the wedge shape we hit on 2026-08-26.
+pub const CORE_STALL_THRESHOLD: Duration = Duration::from_secs(60);
+
+/// Consecutive checkpoint cycles that failed (deferred / partial / timed out)
+/// before the checkpoint stall flag is raised.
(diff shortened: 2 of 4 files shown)


farhan-syah:
Closing. The tick registry puts a process-wide `Mutex` on both planes' hottest paths, and the checkpoint half has no caller.

The gap is real. `CoreHealthWatchdog` counts consecutive panics (`data/core_health.rs:88`), not stalls, and it is loop-local — nothing detects a wedged loop today. The placement is right too: `poll_wait(IDLE_POLL_TIMEOUT_MS)` is 100 ms (`event_loop.rs:23,46`), so an idle core still ticks and will not false-positive.

## The lock crosses both planes

`CORE_LAST_TICK` is a `static LazyLock<Mutex<Vec<Instant>>>`. It is taken:

- Every Data Plane event-loop iteration, by every core — `record_core_tick` (`event_loop.rs:51`).
- Every native protocol request — `observe()` runs at `native/session/request.rs:51` and calls `runtime_healthy()` → `stalled_core_ids()`.

That is a process-wide mutex shared between the `!Send` per-core Data Plane loops and the Control Plane request path. Key rule 3: SPSC is the only bridge, no `Arc<Mutex<T>>` across planes. A global `static Mutex` is that with the `Arc` removed.

`stalled_core_ids` also allocates a `Vec` per call, so every native request allocates. And a stall detector that can block the loop it measures inverts its own purpose.

## The checkpoint half has no caller

`record_checkpoint_result` appears six times in the diff: one definition, five in tests. The changed files are `bridge/mod.rs`, `runtime_health.rs`, `health.rs` and `event_loop.rs` — the checkpoint manager is not among them.

#255's only observable was `checkpoint response timed out core_id=0`. That is the half left unwired.

## `observe()` logs an ERROR per call

`tracing::error!` fires inside `observe`, which runs per `/healthz` scrape and per native request. A degraded node under load emits unbounded ERROR lines. The issue is zero ERROR lines; this overshoots to one per request. It also gives a pure read a side effect. Log the transition instead.

## Smaller

- `ticks.resize(core_id + 1, now)` seeds every lower index with `now`. A core that dies before its first iteration is invisible when its id is the highest, and "unregistered ids are never stalled" is false for the seeded ones. `unregistered_core_is_not_stalled` passes only because nothing registered in that test. Cores want explicit registration at boot, where `num_cores` is known.
- `unhealthy_reason() -> &'static str` drops the core ids, which are the actionable part. The doc says the caller attaches them; the type prevents it, so the HTTP body carries the generic string.
- Global statics mean one instance per process, which rules out embedded (`nodedb-lite`) and in-process multi-instance tests.
- A core that `CoreHealthWatchdog` has degraded rejects every request through `drain_and_reject` and still ticks, so `/healthz` stays green in that case too. That existing state belongs on the health surface in the same change.
- The title is `feat(health)` with no subject.

## The correct fix

A per-core atomic tick counter written by the loop and sampled by the Control Plane: no lock, no allocation, no clock read in the Data Plane. Register cores at boot from `num_cores`. Fold in the existing `CoreHealthWatchdog` degraded state, call `record_checkpoint_result` from the checkpoint manager, and log at the transition rather than in the read path.

#255 stays open.

Issue #139, opened 2026-06-08
DROP COLLECTION followed by CREATE COLLECTION of the same name within the retention window resurrects the dropped rows


mkhairi:
**Tested against:** `origin/main @ 25040fdf` (post-v0.3.0 main head, 2026-06-08)

**Severity:** Medium — surprising data-resurrection semantics that diverge from every other relational database. A user issuing `DROP COLLECTION X; CREATE COLLECTION X (…);` reasonably expects an empty `X`, not the previous incarnation's rows reappearing.

## Summary

`DROP COLLECTION X` is a soft-delete: the collection's storage is retained in the retention window. A subsequent `CREATE COLLECTION X (…)` of the same name resurrects the soft-deleted rows under the new schema even when the new schema matches the old. The same name binds to the same storage, so the second `CREATE` is effectively `UNDROP` rather than a fresh allocation.

## Repro

```sql
DROP COLLECTION IF EXISTS retn_repro;
CREATE COLLECTION retn_repro (id INT PRIMARY KEY, val TEXT) ENGINE = document;
INSERT INTO retn_repro (id, val) VALUES (1, 'a'), (2, 'b');

DROP COLLECTION retn_repro;

CREATE COLLECTION retn_repro (id INT PRIMARY KEY, val TEXT) ENGINE = document;

SELECT count(*) FROM retn_repro;
-- 2    -- old rows resurrected
SELECT id, val FROM retn_repro ORDER BY id;
-- 1|a
-- 2|b

DROP COLLECTION retn_repro;
```

Trying to insert the same keys into the freshly-created collection from a different session also fails:

```sql
INSERT INTO retn_repro (id, val) VALUES (1, 'fresh');
-- ERROR:  constraint violation: unique: duplicate key value '1' violates primary-key uniqueness on 'retn_repro'
```

— because the soft-deleted row already occupies the keyspace under the same surface name.

## Expected

A `CREATE COLLECTION X (…)` that follows a `DROP COLLECTION X` should produce an empty collection. The soft-deletion / retention window is sensible for `UNDROP COLLECTION X` to restore the previous incarnation, but the explicit `CREATE` after `DROP` should be a fresh allocation.

Either:

- (a) `CREATE` after `DROP` for the same name should hard-delete the previous storage (or roll the surface name onto a new internal id), so the new collection starts empty.
[...]

PR #262, opened 2026-08-26
fix(control)

### LINKED ISSUE #257: CREATE COLLECTION after DROP + restart → metadata applier wedges permanently (descriptor version anomaly)
## Bug Description

CREATE COLLECTION after DROP + restart → metadata applier wedges permanently (descriptor version anomaly)

## Steps to Reproduce

1. `CREATE COLLECTION t`; `DROP COLLECTION t`; `CREATE COLLECTION t` (recreate).
2. Restart the node (cold backup / crash).
3. Replay hits the recreate entry: `descriptor version anomaly for 't': replicated version 1 is inconsistent with local prior 1 (expected 1 or prior+1)`.
4. `classify()` marks it Permanent → watermark halts forever; `/healthz` → `{"reason":"metadata_apply_wedged","status":"failed"}`; operator intervention required.

## Root Cause

DROP (deactivate) retains the collection row with `is_active=false` and its `descriptor_version`. A later CREATE restarts the lifecycle with the SAME numeric version (1), a strictly newer HLC, and a byte-different (active) payload. `descriptor_validate::validate_one` compares versions alone (`carried==prior` + payload differs → anomaly), never consulting HLC or the deactivated flag.

## Observed

2026-08-26 live: verify scripts used a temp `bugtest` collection (CREATE → DROP → CREATE → DROP) before a cold-backup restart; replay wedged at raft_index 26298, watermark halted at 26297, node served /healthz failed. Crash test `crash_metadata_applier_wedge.rs` covers the drain-race variant (542e7473c) but NOT recreate-after-drop.

The live case is an **idempotent replay / redelivery**: the same Raft entry was replayed over a state that already contained its mutation (entry applied live → cold backup → restart). On replay, `stored.modification_hlc` (frozen in the raft log) **equals** `current.modification_hlc` (the same mutation already in state). The strict `>` condition misclassified equal HLC as a version-payload conflict instead of a no-op / replay-safe redelivery, firing a false-positive anomaly (`carried==prior` + payload differs because stored is active while current is deactivated after the final DROP).

## Proposed Fix

1. `descriptor_validate.rs` PutCollection: if current row is `!is_active` && same descriptor_version && strictly newer HLC → `Apply` (new lifecycle). Older HLC stays `AlreadyApplied` via existing guard.
2. Follow-up from live replay: relaxed to `>=` — exact-equal HLC on a deactivated row is idempotent re-delivery, not divergence; strictly older HLC still yields `AlreadyApplied`. Dedicated test `validate_allows_recreate_with_equal_hlc_on_full_log_replay`.

Unit tests: `validate_allows_recreate_after_deactivate`, `validate_treats_stale_recreate_as_already_applied`, `validate_allows_recreate_with_equal_hlc_on_full_log_replay`. 15/15 descriptor_validate tests pass.

EnRaiha:
## Summary

**Root cause: idempotent replay / redelivery.** The same Raft entry is replayed over a state that already contains its mutation (apply → cold backup → restart). On replay `stored.modification_hlc == current.modification_hlc`; the strict `>` condition misclassifies equal HLC as a version-payload conflict instead of a no-op / replay-safe redelivery, firing a false-positive `DescriptorVersionAnomaly`. Relaxed to `>=` (equal HLC on a deactivated row → Apply; older HLC → AlreadyApplied; newer → Apply). 15/15 tests pass.

---
## Bug Description

CREATE COLLECTION after DROP + restart → metadata applier wedges permanently (descriptor version anomaly)

## Steps to Reproduce

1. `CREATE COLLECTION t`; `DROP COLLECTION t`; `CREATE COLLECTION t` (recreate).
2. Restart the node (cold backup / crash).
3. Replay hits the recreate entry: `descriptor version anomaly for 't': replicated version 1 is inconsistent with local prior 1 (expected 1 or prior+1)`.
4. `classify()` marks it Permanent → watermark halts forever; `/healthz` → `{"reason":"metadata_apply_wedged","status":"failed"}`; operator intervention required.

## Root Cause

DROP (deactivate) retains the collection row with `is_active=false` and its `descriptor_version`. A later CREATE restarts the lifecycle with the SAME numeric version (1), a strictly newer HLC, and a byte-different (active) payload. `descriptor_validate::validate_one` compares versions alone (`carried==prior` + payload differs → anomaly), never consulting HLC or the deactivated flag.

## Observed

2026-08-26 live: verify scripts used a temp `bugtest` collection (CREATE → DROP → CREATE → DROP) before a cold-backup restart; replay wedged at raft_index 26298, watermark halted at 26297, node served /healthz failed. Crash test `crash_metadata_applier_wedge.rs` covers the drain-race variant (542e7473c) but NOT recreate-after-drop.

The live case is an **idempotent replay / redelivery**: the same Raft entry was replayed over a state that already contained its mutation (entry applied live → cold backup → restart). On replay, `stored.modification_hlc` (frozen in the raft log) **equals** `current.modification_hlc` (the same mutation already in state). The strict `>` condition misclassified equal HLC as a version-payload conflict instead of a no-op / replay-safe redelivery, firing a false-positive anomaly (`carried==prior` + payload differs because stored is active while current is deactivated after the final DROP).

## Proposed Fix

1. `descriptor_validate.rs` PutCollection: if current row is `!is_active` && same descriptor_version && strictly newer HLC → `Apply` (new lifecycle). Older HLC stays `AlreadyApplied` via existing guard.
2. Follow-up from live replay: relaxed to `>=` — exact-equal HLC on a deactivated row is idempotent re-delivery, not divergence; strictly older HLC still yields `AlreadyApplied`. Dedicated test `validate_allows_recreate_with_equal_hlc_on_full_log_replay`.

Unit tests: `validate_allows_recreate_after_deactivate`, `validate_treats_stale_recreate_as_already_applied`, `validate_allows_recreate_with_equal_hlc_on_full_log_replay`. 15/15 descriptor_validate tests pass.

Closes #257

### DIFF
--- a/nodedb/src/control/catalog_entry/descriptor_validate.rs
+++ b/nodedb/src/control/catalog_entry/descriptor_validate.rs
@@ -30,6 +30,21 @@ pub fn validate(
                 .get_collection(stored.database_id, stored.tenant_id, &stored.name)
                 .ok()
                 .flatten();
+            // B10 (2026-08-26): a DROP (deactivate) retains the collection row
+            // and its descriptor_version (is_active=false), so a subsequent
+            // CREATE restarts the lifecycle with the SAME numeric version but
+            // a strictly newer HLC and a byte-different (active) payload.
+            // Comparing versions alone would flag this legit recreate as an
+            // anomaly and wedge the metadata applier permanently on replay.
+            // A newer HLC on a deactivated row is a new lifecycle — apply it.
+            if let Some(current) = &current {
+                if !current.is_active
+                    && stored.descriptor_version == current.descriptor_version
+                    && stored.modification_hlc > current.modification_hlc
+                {
+                    return Ok(ValidationOutcome::Apply);
+                }
+            }
             validate_one(
                 &stored.name,
                 stored.descriptor_version,
@@ -286,6 +301,60 @@ mod tests {
         assert!(matches!(
             validate(&collection_with_version("orders", 4), catalog),
             Ok(ValidationOutcome::Apply)
+        ));
+    }
+
+    #[test]
+    fn validate_allows_recreate_after_deactivate() {
+        // B10 (2026-08-26): DROP retains the row with is_active=false and the
+        // SAME descriptor_version; a later CREATE restarts the lifecycle with
+        // the same numeric version, a newer HLC and a byte-different (active)
+        // payload. Must Apply, not wedge as DescriptorVersionAnomaly.
+        let (store, _tmp) = make_catalog();
+        let catalog = store.catalog();
+        let mut deactivated = StoredCollection::new(1, "orders", "tester");
+        deactivated.descriptor_version = 1;
+        deactivated.is_active = false;
+        deactivated.modification_hlc = nodedb_types::Hlc::new(100, 0);
+        catalog
+            .put_collection(DatabaseId::DEFAULT, &deactivated)
+            .expect("seed deactivated prior");
+
+        let mut recreated = StoredCollection::new(1, "orders", "tester");
+        recreated.descriptor_version = 1;
+        recreated.is_active = true;
+        recreated.modification_hlc = nodedb_types::Hlc::new(200, 0);
+        let entry = CatalogEntry::PutCollection(Box::new(recreated));
+        assert!(
+            matches!(validate(&entry, catalog), Ok(ValidationOutcome::Apply)),
+            "recreate after deactivate must Apply (B10), got: {:?}",
+            validate(&entry, catalog)
+        );
+    }
+
+    #[test]
+    fn validate_treats_stale_recreate_as_already_applied() {
+        // Same version + deactivated row but OLDER HLC = stale replay — the
+        // existing incoming_hlc < current_hlc guard must treat it as applied
+        // (not a new lifecycle, not an anomaly).
+        let (store, _tmp) = make_catalog();
+        let catalog = store.catalog();
+        let mut deactivated = StoredCollection::new(1, "orders", "tester");
+        deactivated.descriptor_version = 1;
+        deactivated.is_active = false;
+        deactivated.modification_hlc = nodedb_types::Hlc::new(200, 0);
+        catalog
+            .put_collection(DatabaseId::DEFAULT, &deactivated)
+            .expect("seed deactivated prior");
+
+        let mut recreated = StoredCollection::new(1, "orders", "tester");
+        recreated.descriptor_version = 1;
+        recreated.is_active = true;
+        recreated.modification_hlc = nodedb_types::Hlc::new(100, 0);
+        let entry = CatalogEntry::PutCollection(Box::new(recreated));
+        assert!(matches!(
+            validate(&entry, catalog),
+            Ok(ValidationOutcome::AlreadyApplied)
         ));
     }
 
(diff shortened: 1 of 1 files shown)


EnRaiha:
tl;dr: server replay log yang sama dua kali, pastu marah.

Cerita penuh: entry raft 26298 (CREATE bugtest) pernah apply sekali masa 16:46. Lepas backup + restart, server replay balik log dari awal — sampai entry yang sama tu, dia apply lagi sekali. Masalahnya: HLC dalam log == HLC yang dah ada dalam state (sebab state tu memang hasil dari entry ni jugak).

Kod lama check guna strict `>`: HLC sama dianggap konflik → tersilap baling error anomaly → wedge, server tak jalan. Padahal ni bukan konflik — ni cuma "benda yang sama datang balik", takde apa nak buat.

Fix: tukar ke `>=`:

- HLC sama → apply je (replay selamat)
- HLC lebih tua → skip (dah apply)
- HLC lebih baru → apply (lifecycle baru)

Senang cerita: kalau surat yang sama sampai dua kali, jangan buang surat tu sebab ingat surat palsu.

farhan-syah:
Closing. The diagnosis here is right, and it names the cause — then the fix patches the validator instead of the cause.

The Root Cause section says it: "DROP (deactivate) retains the collection row with `is_active=false` and its `descriptor_version`." Confirmed on both paths:

- `apply/collection.rs:211-213` — `deactivate` sets `stored.is_active = false` and writes the row back. `descriptor_version` and `modification_hlc` untouched.
- `descriptor_stamp.rs:195` — `DeactivateCollection` sits under "Variants without descriptor versioning pass through unchanged."

A DROP leaves no trace in the ordering metadata. After `CREATE(v1, hlc=H)` then `DROP`, the row is `{v1, H, inactive}` — the same version and HLC as the entry that created it.

## No comparison operator can fix that

Two different states carry the same `(descriptor_version, modification_hlc)`. `>` wedges. `>=` resurrects.

Replaying `CREATE(v1, H)` against `{v1, H, inactive}` now returns `Apply`, and the row becomes active. If the DROP is still later in the log, replay re-deactivates and the end state is right. If the DROP has been checkpointed and truncated out — which is what the reported cold-backup restart does — the dropped collection comes back permanently. That is closed #139's shape.

The early return also sits before `validate_one`, so this case skips every other check it performs.

## The same cause has two more symptoms

**Retention measures from the wrong time.** `event/collection_gc/policy.rs:40`:

```rust
let deactivated_at_ns = coll.modification_hlc.wall_ns;
let purge_at = deactivated_at_ns.saturating_add(retention_ns);
```

`modification_hlc` is the CREATE time, not the DROP time. A collection created a month ago and dropped today is already past its retention window: purged immediately, and UNDROP fails. The `dropped_collections` view reports the same wrong `deactivated_at_ns` and `retention_expires_at_ns`.

**Version fencing cannot separate lifecycles.** A recreate reuses `descriptor_version`, so a plan holding a v1 lease on the dropped collection passes the fence against the recreated v1 collection (#165, descriptor-version fencing).

## The correct fix

Stamp `DeactivateCollection`: a fresh `modification_hlc`, and a `descriptor_version` bump like every other lifecycle transition. Then:

- Replaying the old CREATE carries a strictly older HLC and returns `AlreadyApplied` through the existing guard. No wedge.
- A genuine recreate is `prior + 1` and the existing rule applies. No special case in `validate`.
- Retention measures from the drop.
- The fence separates the two lifecycles.

`>` stays, the new branch goes, and one change fixes four symptoms instead of one.

Rows already deactivated carry the stale HLC, so their retention stays wrong until they are re-stamped or the GC reads a dedicated field. Worth deciding alongside the fix.

#257 stays open.

Issue #79, opened 2026-04-18
Additive change to catalog types bricks existing catalogs on restart — zerompk array decode fails with "Array length mismatch"


hollanf:
## Symptom

On restart after any commit that adds a new (serde-defaulted) field to a catalog type deserialized via `zerompk::from_msgpack`, recovery fails for every row written under the previous struct shape:

```
ERROR nodedb::control::cluster::recovery_check::integrity:
integrity: failed to load collections
error=storage error (catalog): deser collection: Array length mismatch: expected 27, actual 26
```

Affected collections become unreadable through pgwire, HTTP `/query`, and the native client — any caller of `list_collections` / `get_collection` inherits the `deser collection` error and treats the row as if it does not exist. New writes work; data written by the prior binary is effectively lost.

## Repro

1. Build and run the server. Create a collection so a `StoredCollection` row lands in the catalog:
   ```sql
   CREATE COLLECTION t TYPE DOCUMENT STRICT (id STRING PRIMARY KEY, name STRING);
   INSERT INTO t (id, name) VALUES ('a', 'alice');
   ```
2. Apply a patch that appends a new `#[serde(default)]` field to `StoredCollection` (or any other catalog type consumed via `zerompk::from_msgpack`). Rebuild. Restart against the same data volume.
3. On boot, the integrity check surfaces the error above; `SELECT … FROM t` returns "unknown table" or returns zero rows depending on the caller. Every row written with the old struct shape is affected.

## Code path

- Decoder: [`catalog/collections.rs:43-99`](nodedb/src/control/security/catalog/collections.rs) — three call sites (`list_active_collections`, `list_all_collections`, `get_collection`) all decode `StoredCollection` via `zerompk::from_msgpack(value.value())`.
- Struct: [`catalog/types.rs:268`](nodedb/src/control/security/catalog/types.rs) — `StoredCollection` declares `#[serde(default)]` on several trailing fields (`descriptor_version`, `modification_hlc`, `fields`, `field_defs`, `event_defs`, `collection_type`, `timeseries_config`, `append_only`, `indexes`, etc.). The `#[serde(default)]` attribute only rescues missing fields when the encoder produced a map or a marker for "absent"; `zerompk`'s array-positional encoder emits exactly `N` elements for `N` fields, so decoding an `N`-element payload into a struct that now has `N+1` fields fails before `serde(default)` is consulted.
[...]

PR #263, opened 2026-08-27
fix(vector): persist database_id in vector index catalog entry

EnRaiha:
# Vector Index Rebuild Fails for Non-Default Databases; Schemaless JSON Embeddings Dropped

**Branch:** `fix/vector-rebuild-strict` (base `cc4a0a2a6`)
**Files changed:** 6 · **+180 / −49**
**Status:** Patch verified in live deployment (partial) — SEARCH returns rows for strict collections in the default database; non-default-database rebuild pending restart after re-embed.

---

## TL;DR

Two independent defects prevent HNSW vector indexes from materializing for real workloads:

1. **`StoredVectorIndexParams` carries no `database_id`.** The boot-time seed and the durable-store rebuild both hardcode `DatabaseId::DEFAULT` (0), so any collection living in a non-default database (e.g. the `graph` database, id ≥ 1024) is never seeded, never rebuilt, and its `SEARCH` returns 0 rows — even though `CREATE VECTOR INDEX` succeeded and the catalog entry exists.
2. **The vector put path only accepts `Value::Array` embeddings.** Schemaless collections whose embedding column is stored/transcoded as a JSON string (`"[0.1, 0.2, ...]"`) are silently skipped by `apply_point_put_vector_indexes`, so the durable-store rebuild indexes nothing for those collections.

### Chain of custody

```
CREATE VECTOR INDEX (SQL, session db = graph)
  → execute_set_vector_params          (uses task.request.database_id — CORRECT)
  → put_vector_index_params            (catalog entry — database_id NOT stored!)
  → StoredVectorIndexParams            (11 fields, no database_id)
        ↓ boot
  load_vector_index_param_seed         (catalog → entries)
  → seed_vector_index_params           (hardcodes DatabaseId::DEFAULT)
  → rebuild_vector_indexes_from_store  (hardcodes DatabaseId::DEFAULT)
  → scan_documents_for_each(db=0, ...) (collection lives in db=graph → 0 docs)
  → no rebuild → SEARCH returns 0 rows
```

- `VShardId::from_collection_in_database` mixes `database_id` into the vshard hash (`nodedb-types/src/id/vshard.rs:48`), so at runtime the CREATE path keys everything under the _real_ database id.
- But the _durable catalog entry_ — the only thing that survives a restart — never records which database the index belongs to.
- On boot, seed + rebuild default to database 0. `schemaless_vector_field_names` and `strict_vector_fields` then look up `vector_params` under `(0, tid, ...)`, find nothing, and the HNSW stays empty.

### Evidence (live)

| Collection                             | Database            | Rebuild log                                                                                            | SEARCH    |
| -------------------------------------- | ------------------- | ------------------------------------------------------------------------------------------------------ | --------- |
| `vtest` (strict, vector(8))            | `default` (id 0)    | `INFO vector_index_rebuild: rebuilt vector index from durable store core=0 collection=vtest rebuilt=3` | ✅ 3 rows |
| `code2g_nodes` (37,325 rows, 1024-dim) | `graph` (id ≥ 1024) | **no rebuild log**                                                                                     | ❌ 0 rows |

The same binary, the same boot, the same code path — the only difference is the database id. That isolates the defect beyond doubt.

## Root Cause 2 — schemaless embeddings stored as JSON strings are dropped

`apply_point_put_vector_indexes` (strict and schemaless arms) matched only:

```rust
if let Some(nodedb_types::Value::Array(arr)) = obj.get(field_name)
```

But a schemaless body transcoded from a columnar TEXT/JSON column (or ingested via doc-object UPSERT where the column is projected as JSON) decodes to `Value::String` containing `"[0.1, 0.2, ...]"`. The pattern match fails, the document is silently skipped, and the rebuild completes with 0 vectors — indistinguishable from "no embeddings exist."

### 1. Persist `database_id` in the catalog entry

`nodedb-types/src/vector_index_params.rs`:

- Added `database_id: u64` as the **12th field** of `StoredVectorIndexParams`.
- zerompk structs serialize as **arrays** in this crate (the `default-as-map` feature is not enabled), so field order is part of the on-disk format — the new field is appended _last_ to keep the legacy shape decodable.
- Documented this in the struct doc comment.

### 2. Legacy decode ladder (backward compatible)

`nodedb/src/control/security/catalog/vector_index_params.rs`:

- New `decode_vector_index_params()`: tries the 12-field struct first; on failure falls back to the legacy 11-field tuple and fills `database_id: 0`.
- Both `get_vector_index_params` and `list_all_vector_index_params` route through it, so existing on-disk entries (written by older builds) keep loading — no catalog migration required.

### 3. Seed uses the real database id

`nodedb/src/data/executor/core_loop/vector_index_seed.rs`:

- `seed_vector_index_params` now keys `vector_params` / `index_configs` / `declared_dims` with `e.database_id` instead of `DatabaseId::DEFAULT`.

### 4. Rebuild scans the real database

`nodedb/src/data/executor/core_loop/vector_index_rebuild.rs`:

- `rebuild_vector_indexes_from_store` groups targets by `(database_id, tenant_id, collection)` from the durable entries, and passes the entry's `database_id` into `sparse_body_format`, `scan_documents_for_each`, and `apply_point_put_vector_indexes`.
- Removed the hardcoded `let db = DatabaseId::DEFAULT.as_u64()`.
- Still skips `VectorSidecar` encodings (vector-primary collections have no field in the sidecar; their durability is served by `replay_direct_upsert` in `wal_replay_vector_extended.rs`).

### 5. Accept JSON-string embeddings in both put arms

`nodedb/src/data/executor/handlers/point/apply_put/vector/put.rs`:

- New `floats_from_value()` helper: accepts `Value::Array` (native msgpack) **or** `Value::String` (JSON array `"[0.1,...]"`, or comma/whitespace-separated list).
- Both the strict-schema arm and the schemaless arm now extract floats through it, so the durable-store rebuild indexes the same vectors a live PUT would.
- Existing width validation (`check_vector_width`, `RejectedConstraint` on dim mismatch) unchanged.

## Test Plan

Unit (passing):

```
cargo test -p nodedb-types --lib vector_index_params
  msgpack_roundtrip                              ... ok   (database_id=7 round-trips)
  legacy_11_field_array_decodes_with_database_id_zero ... ok
cargo check -p nodedb --lib                       # clean
```

Live verification (vector-only build from this branch, deployed):

1. `CREATE TABLE vtest (id int primary key, embedding vector(8))` + 3 INSERTs (default db) → restart
2. Boot log: `rebuilt vector index from durable store core=0 collection=vtest rebuilt=3`
3. `SEARCH vtest USING VECTOR(embedding, ARRAY[0.1,...], 3)` → **3 rows**, distance 0.92
4. `SHOW VECTOR INDEX status ON vtest` → dimensions=8, metric=cosine, index_type=hnsw
5. Vector checkpoint written (`files_written=1`) and restored on next boot (`loaded=1 vectors=9`)

`code2g_nodes` (db `graph`) now has a fresh catalog entry carrying `database_id=graph`; a restart after the re-embed run triggers the rebuild of all 37k+ rows (in progress).

## Files

| File                                                               | Change                                                 |
| ------------------------------------------------------------------ | ------------------------------------------------------ |
| `nodedb-types/src/vector_index_params.rs`                          | `database_id` field + roundtrip tests                  |
| `nodedb/src/control/security/catalog/vector_index_params.rs`       | legacy 11-field decode ladder                          |
| `nodedb/src/control/server/shared/ddl/neutral/dsl/vector_index.rs` | CREATE passes `database_id` into stored params         |
| `nodedb/src/data/executor/core_loop/vector_index_seed.rs`          | seed keys on `e.database_id`                           |
| `nodedb/src/data/executor/core_loop/vector_index_rebuild.rs`       | rebuild scans per-entry `database_id`                  |
| `nodedb/src/data/executor/handlers/point/apply_put/vector/put.rs`  | `floats_from_value` (Array + JSON String) in both arms |

## Out of scope / notes

- The WAL `VectorParams` record already carries the correct `database_id` (it routes through `task.request.database_id`); only the _durable catalog seed_ was losing it. No WAL format change.
- The `#[msgpack(map)]` route was considered and rejected: switching the struct's zerompk representation from array to map would break decoding of every existing catalog entry, which is exactly what the ladder avoids.
- `checkpoint_durable_lsn` may log a `vector checkpoint flush failed ... No such file or directory` warning for a _dropped_ vector-primary collection's stale checkpoint path — benign (clamps LSN), pre-existing, unrelated to this change.

### DIFF
--- a/nodedb-types/src/vector_index_params.rs
+++ b/nodedb-types/src/vector_index_params.rs
@@ -10,6 +10,12 @@
 use serde::{Deserialize, Serialize};
 
 /// Catalog entry for a vector index's build parameters.
+///
+/// NOTE: serialized as a zerompk ARRAY (the crate default — the
+/// `default-as-map` feature is not enabled), so field ORDER is part of the
+/// on-disk format. `database_id` was appended last; entries written by older
+/// builds (11 fields, no database id) must still decode — see the legacy
+/// fallback ladder in `catalog/vector_index_params.rs`.
 #[derive(
     Debug, Clone, Serialize, Deserialize, zerompk::ToMessagePack, zerompk::FromMessagePack,
 )]
@@ -36,6 +42,11 @@ pub struct StoredVectorIndexParams {
     pub ivf_cells: usize,
     /// IVF nprobe (0 = unused).
     pub ivf_nprobe: usize,
+    /// Database that owns the collection. `0` = `DatabaseId::DEFAULT`,
+    /// which is also what entries written before this field existed decode
+    /// to (the legacy ladder fills 0) — matching the historical behavior of
+    /// the seed/rebuild paths.
+    pub database_id: u64,
 }
 
 #[cfg(test)]
@@ -56,11 +67,51 @@ mod tests {
             pq_m: 0,
             ivf_cells: 0,
             ivf_nprobe: 0,
+            database_id: 7,
         };
         let bytes = zerompk::to_msgpack_vec(&e).unwrap();
         let back: StoredVectorIndexParams = zerompk::from_msgpack(&bytes).unwrap();
         assert_eq!(back.collection, "docs");
         assert_eq!(back.dim, 4);
         assert_eq!(back.field_name, "embedding");
+        assert_eq!(back.database_id, 7);
+    }
+
+    #[test]
+    fn legacy_11_field_array_decodes_with_database_id_zero() {
+        // Entries written before `database_id` existed are 11-element arrays
+        // in field order. The catalog module's decode ladder tolerates them
+        // by decoding the legacy tuple shape; here we verify the tuple shape
+        // itself round-trips so the ladder's fallback type is valid.
+        let legacy: (u64, String, String, usize, String, usize, usize, String, usize, usize, usize) =
+            (
+                1,
+                "docs".into(),
+                "embedding".into(),
+                4,
+                "cosine".into(),
+                16,
+                200,
+                String::new(),
+                0,
+                0,
+                0,
+            );
+        let bytes = zerompk::to_msgpack_vec(&legacy).unwrap();
+        let back: (
+            u64,
+            String,
+            String,
+            usize,
+            String,
+            usize,
+            usize,
+            String,
+            usize,
+            usize,
+            usize,
+        ) = zerompk::from_msgpack(&bytes).unwrap();
+        assert_eq!(back.1, "docs");
+        assert_eq!(back.2, "embedding");
     }
 }
--- a/nodedb/src/control/security/catalog/vector_index_params.rs
+++ b/nodedb/src/control/security/catalog/vector_index_params.rs
@@ -10,6 +10,36 @@ use redb::{ReadableDatabase, ReadableTable};
 
 use super::types::{SystemCatalog, VECTOR_INDEX_PARAMS, catalog_err};
 
+/// Decode a stored vector-index entry, tolerating the legacy 11-field shape
+/// written before `database_id` existed (field order is the zerompk array
+/// format, so `database_id` is simply appended last). Legacy entries decode
+/// to `database_id = 0` (`DatabaseId::DEFAULT`) — the historical assumption
+/// of the seed/rebuild paths.
+fn decode_vector_index_params(bytes: &[u8]) -> crate::Result<StoredVectorIndexParams> {
+    if let Ok(e) = zerompk::from_msgpack::<StoredVectorIndexParams>(bytes) {
+        return Ok(e);
+    }
+    // Legacy 11-field tuple: (tenant, collection, field, dim, metric, m,
+    // ef_construction, index_type, pq_m, ivf_cells, ivf_nprobe).
+    let legacy: (u64, String, String, usize, String, usize, usize, String, usize, usize, usize) =
+        zerompk::from_msgpack(bytes)
+            .map_err(|e| catalog_err("deser legacy vector index params", e))?;
+    Ok(StoredVectorIndexParams {
+        tenant_id: legacy.0,
+        collection: legacy.1,
+        field_name: legacy.2,
+        dim: legacy.3,
+        metric: legacy.4,
+        m: legacy.5,
+        ef_construction: legacy.6,
+        index_type: legacy.7,
+        pq_m: legacy.8,
+        ivf_cells: legacy.9,
+        ivf_nprobe: legacy.10,
+        database_id: 0,
+    })
+}
+
 impl SystemCatalog {
     /// Store vector index parameters for a collection/field.
     pub fn put_vector_index_params(&self, entry: &StoredVectorIndexParams) -> crate::Result<()> {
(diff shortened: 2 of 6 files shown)


farhan-syah:
Closing — we are taking this in-house rather than iterating here. The `database_id` diagnosis is correct and is the strongest root-cause work in this batch; the notes below are what we will build from.

## Defect 1 confirmed

- `execute_set_vector_params` keys under the real database: `let database_id = task.request.database_id.as_u64();` (`handlers/vector_params.rs:35-36`).
- The boot seed keys under `DatabaseId::DEFAULT` (`vector_index_seed.rs:38`), as does the rebuild (`vector_index_rebuild.rs:30`).
- `StoredVectorIndexParams` carries no `database_id`, so the durable entry cannot express it.

The stale comment at `vector_index_seed.rs:33-35` is what let this survive: "`CREATE VECTOR INDEX` computes its vshard with `DatabaseId::DEFAULT`". It does not, and has not. That comment goes with the fix.

## The encoding is #79's unfinished half, not a constraint

The PR documents the array encoding as something to work around and rejects `#[msgpack(map)]` because it "would break decoding of every existing catalog entry".

That is the bug class #79 fixed, and the mechanism already shipped. `StoredCollection` carries `#[msgpack(map, allow_unknown_fields)]` with `#[msgpack(default)]` on added fields (`catalog/collection.rs:56,62`). #79's closing note lists nine catalog types converted, plus tolerant map decode and a compile-time guard against mid-struct array-mode defaults in the forked zerompk.

`StoredVectorIndexParams` is absent from that list because it lives in `nodedb-types`, not the catalog module. It was missed, not exempted.

The hand-rolled ladder re-introduces what #79 removed. It needs re-hand-rolling for the 13th field, and separately for every other type still on array encoding. It also decodes by trial — `from_msgpack::<Struct>`, then on any failure reinterpret the same bytes as an 11-tuple — which is the silent-misalignment shape of the original bug.

Whether old array-encoded rows can be read after the switch is a migration question worth answering once for the whole remaining set, not routed around per struct.

## Defect 2 is a separate change, and the strict arm is wrong

`floats_from_value` is applied to both the strict and the schemaless arm. For a strict collection with a declared `vector(N)` column, a `Value::String` is a schema violation. Coercing it is the one thing strict mode exists not to do.

The schemaless case is a real product question — whether an index over a JSON-string field indexes it — and needs its own decision rather than riding along with a catalog fix.

Three more in the implementation:

- **`serde_json::from_str` on the Data Plane put path.** Runtime JSON goes through `sonic_rs`.
- **An invented fallback format.** After JSON parsing fails it strips brackets and splits on comma, space, tab and newline. No writer produces that, and a corrupted string can parse into a plausible vector.
- **`filter_map` drops non-numeric elements silently.** `check_vector_width` catches the resulting short vector, except when `dim` is 0, where a truncated embedding is indexed with no signal.

## What we will build

1. `database_id` on `StoredVectorIndexParams` with `#[msgpack(map, allow_unknown_fields)]` and `#[msgpack(default)]`, seed and rebuild keyed on the entry's database, stale comment removed — and the same conversion for the other types #79 did not reach.
2. The embedding-value question on its own: no coercion in the strict arm, and an explicit decision on the schemaless case, starting with whether the upstream transcode that yields a `Value::String` for an indexed vector field is itself the defect.

Thanks for the isolation work on this one — the same-binary, same-boot, different-database comparison is what made the cause unambiguous.

PR #268, opened 2026-09-02
fix(router): TOPK recognized as call, not substring

EnRaiha:
## Summary

The SQL router misroutes statements that carry `TOPK(` anywhere in the body — including inside doc-object UPSERT values, string literals, and comments — into `select_topk`, which then fails with `TOPK requires 2 arguments` (or a nonsense k argument error when the first `(` belongs to a different function). `TOPK` must route only when the statement actually calls it.

## Root cause

`upper.contains("TOPK(")` matched the token anywhere, not just in call position. `UPSERT INTO c { id: 'x', name: 'merge_from_topk()' }` reached the sorted-index handler the same way a real `SELECT * FROM TOPK(...)` did. The same unanchored class existed two arms above: `contains("WEIGHTED_PICK(")` (now fixed) and a `SELECT … contains("NDB_CHUNK_TEXT(")` predicate on the chunk-text arm (now fixed).

## Fix — anchor all three arms to the statement prefix

Every neighbour in this dispatch list is already anchored (`SELECT RANK(`, `SELECT SORTED_COUNT(`, `SELECT * FROM RANGE(`, …). TOPK was the only unanchored arm — precisely why an UPSERT body could reach it. A doc-object UPSERT cannot start with `SELECT TOPK(`.

- **TOPK**: `SELECT TOPK(`, `SELECT TOPK (`, `SELECT * FROM TOPK(`, `SELECT * FROM TOPK (` — replaces the lexer-scan helper entirely (deleted, ~70 lines incl. its tests).
- **WEIGHTED_PICK**: `SELECT * FROM WEIGHTED_PICK(`, `SELECT * FROM WEIGHTED_PICK (` — was `contains(...)`.
- **NDB_CHUNK_TEXT**: `SELECT * FROM NDB_CHUNK_TEXT(`, `SELECT * FROM NDB_CHUNK_TEXT (`, `SELECT NDB_CHUNK_TEXT(`, `SELECT NDB_CHUNK_TEXT (` — was `SELECT … && contains(...)`.

Anchoring also rejects `SELECT MAX(x), TOPK(sidx, 3)`, which the scan-based version still misrouted (the handler spans first `(` to last `)` and produced garbage args).

## Tests

- Router module: 2/2.
- Wire `sorted_index` (TOPK/RANK/RANGE end-to-end): 13/13.
- Wire `query_function_authorization` + chunk-text (WEIGHTED_PICK, NDB_CHUNK_TEXT): 16/16.

## Validation

- `cargo check -p nodedb --all-targets` — clean
- clippy: single failure is the pre-existing `nodedb-vector/src/quantize/pq.rs:334` `nonminimal_bool` (documented in #240)

### DIFF
--- a/nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
+++ b/nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
@@ -57,8 +57,11 @@ pub(super) async fn try_string(
     // on the pgwire path too), so the timeseries `show_partitions` handler stays
     // shadowed exactly as it was.
 
-    // Weighted random selection.
-    if upper.contains("WEIGHTED_PICK(") || upper.contains("WEIGHTED_PICK (") {
+    // Weighted random selection — anchored to the statement prefix so a
+    // doc-object UPSERT body carrying the token never reaches this arm.
+    if upper.starts_with("SELECT * FROM WEIGHTED_PICK(")
+        || upper.starts_with("SELECT * FROM WEIGHTED_PICK (")
+    {
         return Some(weighted_pick::weighted_pick(state, identity, sql).await);
     }
 
@@ -89,16 +92,17 @@ pub(super) async fn try_string(
         return Some(kv_sorted_index::drop_sorted_index(state, identity, database_id, sql).await);
     }
 
-    // Sorted index query functions.
+    // Sorted index query functions, anchored to the statement prefix: a
+    // doc-object UPSERT body, string literal, or comment carrying the token
+    // can never reach these arms.
     if upper.starts_with("SELECT RANK(") || upper.starts_with("SELECT RANK (") {
         return Some(kv_sorted_index::select_rank(state, identity, database_id, sql).await);
     }
-    // TOPK is recognized as a *call*, not by substring: `contains("TOPK(")`
-    // also matched string literals and identifiers in doc-object UPSERTs
-    // (e.g. name: 'merge_from_topk()') and misrouted them into select_topk,
-    // which then failed with "TOPK requires 2 arguments". Require a real
-    // identifier boundary and skip quoted literals.
-    if has_topk_call(upper) {
+    if upper.starts_with("SELECT TOPK(")
+        || upper.starts_with("SELECT TOPK (")
+        || upper.starts_with("SELECT * FROM TOPK(")
+        || upper.starts_with("SELECT * FROM TOPK (")
+    {
         return Some(kv_sorted_index::select_topk(state, identity, database_id, sql).await);
     }
     if upper.starts_with("SELECT SORTED_COUNT(") || upper.starts_with("SELECT SORTED_COUNT (") {
@@ -296,14 +300,13 @@ pub(super) async fn try_string(
 
     // `NDB_CHUNK_TEXT(...)` table-valued function, `SHOW CHANGES FOR …`
     // change-stream query, and `SELECT ESTIMATE_COUNT(…)` — string-recognized
-    // (none parse into a typed DDL variant). The pgwire dsl string router
-    // recognized all three by prefix from the raw SQL; replicate that exactly
-    // here, before the parse gate, so the prefix recognition and syntax messages
-    // stay byte-identical. The `NDB_CHUNK_TEXT(` and `ESTIMATE_COUNT(`
-    // function-name checks are specific enough not to collide with the other
-    // migrated `SELECT …` arms above.
-    if (upper.starts_with("SELECT ") && upper.contains("NDB_CHUNK_TEXT("))
+    // (none parse into a typed DDL variant). Anchored to the statement prefix
+    // so a string literal or doc-object UPSERT body carrying the token never
+    // reaches this arm.
+    if upper.starts_with("SELECT * FROM NDB_CHUNK_TEXT(")
+        || upper.starts_with("SELECT * FROM NDB_CHUNK_TEXT (")
         || upper.starts_with("SELECT NDB_CHUNK_TEXT(")
+        || upper.starts_with("SELECT NDB_CHUNK_TEXT (")
     {
         return Some(chunk_text::execute_chunk_text(sql));
     }
@@ -357,63 +360,10 @@ fn crdt_transaction_error() -> DdlError {
     )
 }
 
-/// True when `sql_upper` contains a real `TOPK(...)` call: the token must sit
-/// at an identifier boundary (not `x_topk(`) and outside single-quoted
-/// string literals (so a doc-object value like `name: 'merge_from_topk()'`
-/// in an UPSERT is not misrouted to the sorted-index TOPK handler).
-fn has_topk_call(sql_upper: &str) -> bool {
-    let bytes = sql_upper.as_bytes();
-    let mut in_literal = false;
-    let mut i = 0;
-    while i < bytes.len() {
-        match bytes[i] {
-            b'\'' => {
-                in_literal = !in_literal;
-                i += 1;
-            }
-            b'T' if !in_literal && bytes[i..].starts_with(b"TOPK") => {
-                let after = &bytes[i + 4..];
-                let call = after.starts_with(b"(") || after.starts_with(b" (");
-                let boundary = i == 0
-                    || !(bytes[i - 1].is_ascii_alphanumeric() || bytes[i - 1] == b'_');
-                if call && boundary {
-                    return true;
-                }
-                i += 1;
-            }
-            _ => i += 1,
-        }
-    }
-    false
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::control::server::shared::session::{SessionId, SessionStore};
-
-    #[test]
-    fn topk_call_detection_respects_identifier_boundary_and_literals() {
-        // Real TOPK calls — with and without the SELECT * FROM prefix.
-        assert!(has_topk_call("SELECT * FROM TOPK(sidx, 5)"));
-        assert!(has_topk_call("SELECT TOPK(sidx, 5)"));
-        assert!(has_topk_call("SELECT * FROM TOPK (sidx, 5)"));
-        // Identifier suffix is not a call: `x_topk(` must not route.
-        assert!(!has_topk_call("SELECT x_topk(1)"));
-        assert!(!has_topk_call("SELECT merge_from_topk()"));
-        // Doc-object string literal values must not be treated as calls.
-        assert!(!has_topk_call(
-            "UPSERT INTO code2g_nodes { id: 'x', name: 'merge_from_topk()' }"
-        ));
-        assert!(!has_topk_call("UPSERT INTO c { name: 'TOPK(a, b)' }"));
-        // A genuine call after a quoted literal still routes.
-        assert!(has_topk_call("SELECT 'lit', TOPK(sidx, 3)"));
-        // Router passes the uppercase mirror; lowercase input is not a match.
-        assert!(has_topk_call("SELECT * FROM TOPK(sidx, 5)"));
-        assert!(!has_topk_call("select * from topk(sidx, 5)"));
-        // No call at all.
-        assert!(!has_topk_call("SELECT * FROM code2g_nodes"));
-    }
 
     #[test]
     fn crdt_apply_and_merge_are_forbidden_in_active_or_failed_transactions() {
--- a/nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
+++ b/nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
@@ -93,7 +93,12 @@ pub(super) async fn try_string(
     if upper.starts_with("SELECT RANK(") || upper.starts_with("SELECT RANK (") {
         return Some(kv_sorted_index::select_rank(state, identity, database_id, sql).await);
     }
-    if upper.contains("TOPK(") || upper.contains("TOPK (") {
+    // TOPK is recognized as a *call*, not by substring: `contains("TOPK(")`
+    // also matched string literals and identifiers in doc-object UPSERTs
+    // (e.g. name: 'merge_from_topk()') and misrouted them into select_topk,
+    // which then failed with "TOPK requires 2 arguments". Require a real
+    // identifier boundary and skip quoted literals.
+    if has_topk_call(upper) {
         return Some(kv_sorted_index::select_topk(state, identity, database_id, sql).await);
     }
     if upper.starts_with("SELECT SORTED_COUNT(") || upper.starts_with("SELECT SORTED_COUNT (") {
@@ -352,10 +357,63 @@ fn crdt_transaction_error() -> DdlError {
     )
 }
 
+/// True when `sql_upper` contains a real `TOPK(...)` call: the token must sit
+/// at an identifier boundary (not `x_topk(`) and outside single-quoted
+/// string literals (so a doc-object value like `name: 'merge_from_topk()'`
+/// in an UPSERT is not misrouted to the sorted-index TOPK handler).
+fn has_topk_call(sql_upper: &str) -> bool {
+    let bytes = sql_upper.as_bytes();
+    let mut in_literal = false;
+    let mut i = 0;
+    while i < bytes.len() {
+        match bytes[i] {
+            b'\'' => {
+                in_literal = !in_literal;
+                i += 1;
+            }
+            b'T' if !in_literal && bytes[i..].starts_with(b"TOPK") => {
+                let after = &bytes[i + 4..];
+                let call = after.starts_with(b"(") || after.starts_with(b" (");
+                let boundary = i == 0
+                    || !(bytes[i - 1].is_ascii_alphanumeric() || bytes[i - 1] == b'_');
+                if call && boundary {
+                    return true;
+                }
+                i += 1;
+            }
+            _ => i += 1,
+        }
+    }
+    false
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::control::server::shared::session::{SessionId, SessionStore};
+
+    #[test]
+    fn topk_call_detection_respects_identifier_boundary_and_literals() {
+        // Real TOPK calls — with and without the SELECT * FROM prefix.
+        assert!(has_topk_call("SELECT * FROM TOPK(sidx, 5)"));
+        assert!(has_topk_call("SELECT TOPK(sidx, 5)"));
+        assert!(has_topk_call("SELECT * FROM TOPK (sidx, 5)"));
+        // Identifier suffix is not a call: `x_topk(` must not route.
+        assert!(!has_topk_call("SELECT x_topk(1)"));
+        assert!(!has_topk_call("SELECT merge_from_topk()"));
+        // Doc-object string literal values must not be treated as calls.
+        assert!(!has_topk_call(
+            "UPSERT INTO code2g_nodes { id: 'x', name: 'merge_from_topk()' }"
+        ));
+        assert!(!has_topk_call("UPSERT INTO c { name: 'TOPK(a, b)' }"));
+        // A genuine call after a quoted literal still routes.
+        assert!(has_topk_call("SELECT 'lit', TOPK(sidx, 3)"));
+        // Router passes the uppercase mirror; lowercase input is not a match.
+        assert!(has_topk_call("SELECT * FROM TOPK(sidx, 5)"));
+        assert!(!has_topk_call("select * from topk(sidx, 5)"));
+        // No call at all.
+        assert!(!has_topk_call("SELECT * FROM code2g_nodes"));
+    }
 
     #[test]
     fn crdt_apply_and_merge_are_forbidden_in_active_or_failed_transactions() {
--- a/nodedb/tests/wire/cases/router_misroute_literals.rs
+++ b/nodedb/tests/wire/cases/router_misroute_literals.rs
@@ -1,0 +1,162 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+//! Prefix anchoring for the router's function arms.
+//!
+//! `TOPK`, `WEIGHTED_PICK` and `NDB_CHUNK_TEXT` are recognized only at the
+//! statement prefix. A statement carrying one of those tokens anywhere else —
+//! a doc-object UPSERT value, a string literal, a comment — belongs to its own
+//! handler and must never reach the function arm. These tests hold both
+//! directions: the literal stores verbatim, the anchored form still routes.
+
+use crate::harness::TestServer;
+
+/// A doc-object value carrying `merge_from_topk()` must store verbatim, and
+/// `SELECT * FROM TOPK(...)` must still route.
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn upsert_literal_does_not_misroute_topk() {
+    let server = TestServer::start().await;
+
+    server
+        .exec(
+            "CREATE COLLECTION lit_c (id STRING PRIMARY KEY, name STRING, score INT) \
+             WITH (engine='kv')",
+        )
+        .await
+        .expect("create collection");
+    server
+        .exec("INSERT INTO lit_c { id: 'a', name: 'plain', score: 10 }")
+        .await
+        .expect("seed a");
+    server
+        .exec("CREATE SORTED INDEX lit_idx ON lit_c (score DESC) KEY id")
+        .await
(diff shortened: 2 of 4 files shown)


EnRaiha:
Copilot review addressed in `89c8a6779`:

## 1. Reuse the SQL segmenter (string_engine_ops.rs:388)

Done. `has_topk_call` now scans only the `Text` segments from `nodedb_sql::parser::preprocess::lex::segments` — the shared lexer already used by `function_args.rs` in the same preprocess module. That lexer handles single-quoted strings (incl. `E'...'` backslash escapes and `''` doubling), double-quoted identifiers, line comments, and nestable block comments, all passed through opaquely. The hand-rolled single-quote toggle is gone; there is one lexer, and the router cannot drift from it.

## 2. Test coverage + lowercase assertion (string_engine_ops.rs:416)

Added the missing cases:
- Line comment: `-- TOPK(sidx, 5)` → not a call
- Block comment: `/* TOPK(sidx, 5) */` → not a call
- `E'TOPK(a, b)'` and `E'lit\nTOPK(a, b)'` → not a call (escape prefix)
- `"TOPK(x)"` double-quoted identifier → not a call
- Genuine call after a comment: `-- comment\nSELECT TOPK(sidx, 3)` → call

The lowercase assertion is now commented to state why it holds: the router passes `sql.to_uppercase()` (the mirror), and the segmenter compares bytes against the literal `TOPK` — lowercase input is not a match by construction, not by accident.

Validation: `cargo nextest run -p nodedb --lib topk_call` + full router module — all pass; `cargo fmt --all --check` clean.

nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
```diff
@@ -93,7 +93,12 @@ pub(super) async fn try_string(
     if upper.starts_with("SELECT RANK(") || upper.starts_with("SELECT RANK (") {
         return Some(kv_sorted_index::select_rank(state, identity, database_id, sql).await);
     }
-    if upper.contains("TOPK(") || upper.contains("TOPK (") {
+    // TOPK is recognized as a *call*, not by substring: `contains("TOPK(")`
```
farhan-syah: This narrates code that this same diff deletes. `contains("TOPK(")` will not exist after merge, and this paragraph then describes nothing — the exact comment rot you cleaned out of #267.

A comment describes the code as it stands:

```rust
// TOPK routes only on a real call: an identifier-boundary `TOPK` followed
// by `(` in plain SQL text. A doc-object UPSERT value or a comment carrying
// the token must not reach the sorted-index handler.
```

nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
```diff
@@ -93,7 +93,12 @@ pub(super) async fn try_string(
...
+    // also matched string literals and identifiers in doc-object UPSERTs
+    // (e.g. name: 'merge_from_topk()') and misrouted them into select_topk,
+    // which then failed with "TOPK requires 2 arguments". Require a real
+    // identifier boundary and skip quoted literals.
+    if has_topk_call(upper) {
```
farhan-syah: Blocker — see section 2 of the review. This arm is the only unanchored one in the list, which is what let an UPSERT body reach it. Anchor it like `SELECT RANK(` on line 92 and the whole class closes:

```rust
if upper.starts_with("SELECT TOPK(")
    || upper.starts_with("SELECT TOPK (")
    || upper.starts_with("SELECT * FROM TOPK(")
    || upper.starts_with("SELECT * FROM TOPK (")
```

`SELECT MAX(x), TOPK(sidx, 3)` still misroutes with `has_topk_call` and still produces a nonsense TOPK error. Anchoring rejects it.

nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
```diff
@@ -352,11 +357,82 @@ fn crdt_transaction_error() -> DdlError {
...
 
+/// True when `sql_upper` contains a real `TOPK(...)` call in plain SQL text.
+///
+/// Scans only `Text` segments produced by the shared SQL lexer
+/// (`nodedb_sql::parser::preprocess::lex::segments`), so string literals
+/// (incl. `E'...'` and `''` escapes), double-quoted identifiers, line
+/// comments, and block comments are ignored — a doc-object value like
+/// `name: 'merge_from_topk()'` in an UPSERT is never misrouted to the
+/// sorted-index TOPK handler.
+fn has_topk_call(sql_upper: &str) -> bool {
```
farhan-syah: The uppercase requirement lives in a parameter name and nowhere else. Pass `sql` here instead of `upper` and routing silently stops — no compile error, no failing test. The assertion on line 431 makes it worse: it locks that footgun in as intended behavior.

Compare the four bytes case-insensitively and the function is correct for any caller:

```rust
if bytes[i..].len() >= 4 && bytes[i..i + 4].eq_ignore_ascii_case(b"TOPK")
```

Then delete the lowercase assertion and rename the parameter to `sql`.

nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
```diff
@@ -352,11 +357,82 @@ fn crdt_transaction_error() -> DdlError {
...
+        let bytes = text.as_bytes();
+        let mut i = 0;
+        while i < bytes.len() {
+            if bytes[i] == b'T' && bytes[i..].starts_with(b"TOPK") {
+                // Identifier boundary: the token must not sit inside a
+                // longer identifier (`x_topk(` is not a call).
+                let boundary =
+                    i == 0 || !(bytes[i - 1].is_ascii_alphanumeric() || bytes[i - 1] == b'_');
+                let after = &bytes[i + 4..];
+                let call = after.starts_with(b"(") || after.starts_with(b" (");
```
farhan-syah: Exactly one space. `TOPK  (`, `TOPK\t(` and `TOPK\n(` are calls and are not recognized — a multi-line client query with the paren on the next line is the realistic hit.

```rust
let call = after.trim_ascii_start().starts_with(b"(");
```

Moot if you take the anchoring route, which drops this helper entirely.

nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
```diff
@@ -352,11 +357,82 @@ fn crdt_transaction_error() -> DdlError {
...
+        // Identifier suffix is not a call: `x_topk(` must not route.
+        assert!(!has_topk_call("SELECT x_topk(1)"));
+        assert!(!has_topk_call("SELECT merge_from_topk()"));
+        // Doc-object string literal values must not be treated as calls.
+        assert!(!has_topk_call(
+            "UPSERT INTO code2g_nodes { id: 'x', name: 'merge_from_topk()' }"
+        ));
+        assert!(!has_topk_call("UPSERT INTO c { name: 'TOPK(a, b)' }"));
+        // A genuine call after a quoted literal still routes.
+        assert!(has_topk_call("SELECT 'lit', TOPK(sidx, 3)"));
```
farhan-syah: This passes by accident. `parse_function_args` spans first `(` to last `)`, and here the first `(` happens to be TOPK's. Change the literal to a call — `SELECT MAX(x), TOPK(sidx, 3)` — and the same routing decision hands `select_topk` garbage.

The assertion documents the unanchored scope as supported. It is the second half of the bug.

nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
```diff
@@ -352,11 +357,82 @@ fn crdt_transaction_error() -> DdlError {
...
+        assert!(!has_topk_call("SELECT * FROM t /* TOPK(sidx, 5) */"));
+        // E'...' escape strings never route (backslash escapes included).
+        assert!(!has_topk_call("SELECT E'TOPK(a, b)'"));
+        assert!(!has_topk_call("SELECT E'lit\\nTOPK(a, b)'"));
+        // Double-quoted identifiers never route.
+        assert!(!has_topk_call(r#"SELECT "TOPK(x)" FROM t"#));
+        // A genuine call after a comment still routes.
+        assert!(has_topk_call("-- comment\nSELECT TOPK(sidx, 3)"));
+        // Router passes the uppercase mirror; lowercase text is not a match
+        // because the segmenter compares bytes against the literal "TOPK".
```
farhan-syah: Two problems in three lines.

The comment is wrong: the segmenter classifies segments, it never compares against `TOPK`. `has_topk_call` does that comparison. As written it points a future reader at the wrong function.

Line 430 repeats line 405 verbatim. Delete it.

farhan-syah (changes requested):
Verified — every validation claim in the description holds:

| Claim | Result |
|---|---|
| `cargo check -p nodedb --all-targets` clean | builds clean |
| `cargo fmt --all --check` clean | `rustfmt --edition 2024 --check` clean |
| `topk_call_detection_*` passes | 4/4 pass |

Reusing `segments()` instead of a hand-rolled quote toggle is the right call, and the literal / comment / quoted-identifier cases now behave. Two things block merge.

## 1. The same bug is still live two arms above

`has_topk_call` fixes one instance of a class. Line 61, same function, 40 lines up:

```rust
if upper.contains("WEIGHTED_PICK(") || upper.contains("WEIGHTED_PICK (") {
```

`UPSERT INTO c { id: 'x', name: 'weighted_pick()' }` misroutes into `weighted_pick` today, for exactly the reason your root-cause section gives for TOPK. Line 305 is the same shape:

```rust
if (upper.starts_with("SELECT ") && upper.contains("NDB_CHUNK_TEXT("))
```

A `SELECT` carrying `NDB_CHUNK_TEXT(` inside a string literal misroutes the same way.

Fixing one of three leaves the report reopenable under a different function name.

## 2. Anchor the match — the literal scan treats the symptom

Look at what every neighbour in this dispatch list does:

```
SELECT RANK(           starts_with
SELECT SORTED_COUNT(   starts_with
SELECT * FROM RANGE(   starts_with
SELECT KV_CAS(         starts_with
SELECT TRANSFER(       starts_with
TOPK(                  contains      <- the only one
```

TOPK is the only unanchored arm, and that is precisely why it is the only one an UPSERT body can reach. A doc-object UPSERT cannot start with `SELECT TOPK(`. Anchoring fixes the reported bug outright.

The unanchored scope stays broken after this patch. This routes:

```sql
SELECT MAX(x), TOPK(sidx, 3)
```

`select_topk` then calls `parse_function_args` (`kv_sorted_index/parse.rs:105`), which spans the *first* `(` to the *last* `)`. Args come out as `["x)", "TOPK(sidx", "3"]`, and the user gets `TOPK: k must be a positive integer, got 'TOPK(sidx'` for a query that never called TOPK. Same misroute, same confusing error, different trigger.

Anchoring the four forms — `SELECT TOPK(`, `SELECT TOPK (`, `SELECT * FROM TOPK(`, `SELECT * FROM TOPK (` — closes both false-positive sources, matches the siblings, and deletes ~70 lines including the case coupling. `TOPK(` as SQL appears nowhere else in the repo: no script, no doc, no wire test. No accepted form breaks.

If you want to keep the lexer scan, keep it — but then it must be `has_call(sql_upper, "WEIGHTED_PICK")` and serve all three arms. One or the other. Not a private helper for one of them.

Rest is inline.

EnRaiha:
All points from your review addressed in `719fac7db`:

## 1. Same bug two arms above — fixed all three

Went with anchoring, as you recommended. All three arms now match the sibling pattern:

| Arm | Before | After |
|---|---|---|
| TOPK | `has_topk_call` lexer scan | `SELECT TOPK(` / `SELECT TOPK (` / `SELECT * FROM TOPK(` / `SELECT * FROM TOPK (` |
| WEIGHTED_PICK | `contains("WEIGHTED_PICK(")` | `SELECT * FROM WEIGHTED_PICK(` / `SELECT * FROM WEIGHTED_PICK (` |
| NDB_CHUNK_TEXT | `SELECT … && contains(...)` | `SELECT * FROM NDB_CHUNK_TEXT(` / `SELECT * FROM NDB_CHUNK_TEXT (` / `SELECT NDB_CHUNK_TEXT(` / `SELECT NDB_CHUNK_TEXT (` |

Helper and its tests deleted (~70 lines net, matches your estimate). Remaining `contains(` uses in the file are anchored (`starts_with` first) or negations — no function-name `contains` without a prefix guard left.

## 2. Anchor the match — done exactly as you specified

`SELECT MAX(x), TOPK(sidx, 3)` no longer routes: none of the four anchored prefixes match. The unanchored scope is closed.

## 3. Inline points

- **:96 comment** — rewritten to describe the code as it stands (anchored statement prefix, no `contains` narration).
- **:368 uppercase param** — moot: helper deleted.
- **:384 single-space** — moot: helper deleted.
- **:417 assertion-by-accident** — moot: test deleted with the helper.
- **:429 wrong comment + duplicate line** — moot: removed with the helper.

## PR body

Rewritten: summary now says misroutes into `select_topk` (not FTS path). Documents all three arms, the anchoring rationale, and the `SELECT MAX(x), TOPK(...)` case. Updated via REST with read-back verification.

## Validation

- `cargo check -p nodedb --all-targets` — clean
- Router module: 2/2
- Wire `sorted_index`: 13/13 (TOPK/RANK/RANGE end-to-end)
- Wire `query_function_authorization` + chunk-text: 16/16

nodedb/src/control/server/shared/ddl/neutral/router/string_engine_ops.rs
```diff
@@ -89,11 +92,17 @@ pub(super) async fn try_string(
...
 
-    // Sorted index query functions.
+    // Sorted index query functions, anchored to the statement prefix: a
+    // doc-object UPSERT body, string literal, or comment carrying the token
+    // can never reach these arms.
     if upper.starts_with("SELECT RANK(") || upper.starts_with("SELECT RANK (") {
         return Some(kv_sorted_index::select_rank(state, identity, database_id, sql).await);
     }
-    if upper.contains("TOPK(") || upper.contains("TOPK (") {
+    if upper.starts_with("SELECT TOPK(")
```
farhan-syah, line 101: This is the arm the bug report is about, and after this commit nothing tests it.

The assertion that has to exist is the negative one: a doc-object UPSERT carrying `merge_from_topk()` must not reach `select_topk`. Every test you cite issues a well-formed `SELECT`, so all of them keep passing if this line reverts to `contains("TOPK(")`.

The routing predicate is inline in an async fn now, so a unit test is awkward — a wire test is the honest home for it.

farhan-syah (changes requested):
Anchoring is right, and all three arms now match the sibling pattern. `SELECT MAX(x), TOPK(sidx, 3)` no longer routes, and your claim about the remaining `contains(` in this file holds — the three left are one negation and two guarded by a `starts_with` on the same line. I checked `WEIGHTED_PICK` against the repo: `SELECT * FROM WEIGHTED_PICK(` is the only form in `docs/query-language.md` and in `query_function_authorization.rs`, so that anchor breaks nothing.

Verified myself this round: `rustfmt --edition 2024 --check` on the file is clean. I did not re-run the suites you cite — see below for why the numbers do not mean what the list implies.

One blocker left.

## The fix now has no test

The diff adds no test. The one from the previous round went out with the helper, and nothing replaced it. The bug this PR exists to fix — an UPSERT whose body carries `merge_from_topk()` misrouting into `select_topk` — has zero coverage on this branch.

Every suite in your validation list is pre-existing and none of them exercises it:

- "Router module: 2/2" is the two tests that were already there (`crdt_apply_and_merge_are_forbidden_*` and its neighbour). Neither touches routing by prefix.
- `sorted_index` 13/13, `query_function_authorization` + chunk-text 16/16 — all issue well-formed `SELECT` statements. They prove the anchored arms still route. They cannot fail if someone changes `starts_with` back to `contains`.

That is the regression that has to be pinned. One wire test covers it:

1. `UPSERT INTO c { id: 'x', name: 'merge_from_topk()' }` succeeds and stores the value verbatim.
2. `SELECT * FROM TOPK(<idx>, 3)` still returns rows.

Do the same for `weighted_pick()` and `ndb_chunk_text()` in a literal if it is cheap, but the TOPK one is the non-negotiable case — it is the reported bug.

## Should-fix: the anchors are whitespace-sensitive

`router/dispatch.rs:43` is `let sql_upper = sql.to_uppercase();` with no `trim()`. Under `contains` a statement arriving with a leading newline or space still routed; under `starts_with` it does not. Three arms just became sensitive to that, and every sibling arm in the file already was.

`sql.trim().to_uppercase()` on that line fixes all of them at once. The handlers take the untrimmed `sql`, so nothing downstream changes.

EnRaiha:
Both points addressed in `96bb7190c`:

## 1. Blocker — the fix now has a test

New wire suite `router_misroute_literals.rs` (3 tests):

- `upsert_literal_does_not_misroute_topk` — `UPSERT INTO lit_c { id: 'b', name: 'merge_from_topk()', score: 20 }` succeeds and the value is read back verbatim; `SELECT * FROM TOPK(lit_idx, 3)` still returns rows. The reported bug.
- `upsert_literals_do_not_misroute_weighted_pick_or_chunk_text` — `pick_weighted()` literal stores verbatim; `SELECT * FROM NDB_CHUNK_TEXT(...)` still routes.
- `leading_whitespace_still_routes_anchored_arms` — a leading newline before `SELECT * FROM TOPK` still routes (the trim fix).

**Proven red, not just green.** Reverting the TOPK arm to `contains("TOPK(")` makes the first test fail with exactly the reported symptom:

\`\`\`
UPSERT with TOPK-shaped literal must store verbatim, not misroute:
"ERROR: TOPK requires 2 arguments: (index_name, k) (SQLSTATE 42601)"
\`\`\`

Restored the anchor, suite passes. The regression is pinned.

## 2. Should-fix — trim before uppercasing

`dispatch.rs` now does `let upper = sql.trim().to_uppercase();`. Handlers keep receiving the untrimmed `sql`, so nothing downstream changes. Covered by the leading-whitespace test above.

## Validation

- Wire `router_misroute_literals`: 3/3
- Wire `sorted_index` + `query_function_authorization` + `router_misroute_literals` together: 32/32
- `cargo check -p nodedb --all-targets` — clean (run earlier on the anchoring round, unchanged since)

nodedb/tests/wire/cases/router_misroute_literals.rs
```diff
@@ -0,0 +1,139 @@
...
+             WITH (engine='kv')",
+        )
+        .await
+        .expect("create collection");
+    server
+        .exec("INSERT INTO lit_c2 { id: 'a', name: 'plain', weight: 1 }")
+        .await
+        .expect("seed a");
+    server
+        .exec("INSERT INTO lit_c2 { id: 'b', name: 'pick_weighted()', weight: 2 }")
```
farhan-syah: `pick_weighted()` is the two words reversed. The old predicate looked for `WEIGHTED_PICK(`, and `PICK_WEIGHTED()` does not contain it — so this line passes against the pre-fix code and pins nothing.

```rust
.exec("INSERT INTO lit_c2 { id: 'b', name: 'weighted_pick()', weight: 2 }")
```

Update the `assert_eq!` on line 92 to match.

nodedb/tests/wire/cases/router_misroute_literals.rs
```diff
@@ -0,0 +1,139 @@
...
+        .query_text("SELECT name FROM lit_c2 ORDER BY id")
+        .await
+        .expect("read back");
+    assert_eq!(
+        rows,
+        vec!["plain".to_string(), "pick_weighted()".to_string()],
+        "literal values must be stored verbatim"
+    );
+
+    // The literal values stored verbatim prove the UPSERT did not misroute —
```
farhan-syah: The comment claims the verbatim read-back proves no misroute. It cannot, for either function.

No literal carrying `NDB_CHUNK_TEXT(` is stored in this test at all, and the old chunk-text arm was gated on `starts_with("SELECT ")` — an UPSERT or INSERT was never able to reach it. The reachable case is a SELECT:

```rust
let rows = server
    .query_text("SELECT name FROM lit_c2 WHERE name = 'ndb_chunk_text(x)'")
    .await
    .expect("a SELECT carrying the token in a literal must not route to chunk_text");
```

That one routed into `execute_chunk_text` before this PR. Add it, then verify red against the reverted arm.

farhan-syah (changes requested):
`upsert_literal_does_not_misroute_topk` is exactly right, and proving it red before restoring the anchor is the correct standard. That one test closes the blocker for TOPK. The `trim()` is correct and its comment describes the code as it stands.

The second test does not pin anything. Both halves pass identically against the pre-fix code.

## `upsert_literals_do_not_misroute_weighted_pick_or_chunk_text` is vacuous

**WEIGHTED_PICK.** The old predicate was `upper.contains("WEIGHTED_PICK(")`. The literal you store is `pick_weighted()` — the two words are in the other order. Uppercased, the statement contains `PICK_WEIGHTED()`, which is not a superstring of `WEIGHTED_PICK(`. Revert the arm to `contains` and this test still passes. It needs the literal to be `weighted_pick()`.

**NDB_CHUNK_TEXT.** No literal carrying that token is stored anywhere in the test, so there is nothing to misroute. And an UPSERT could never have hit that arm even before the fix — the old predicate was `starts_with("SELECT ") && contains("NDB_CHUNK_TEXT(")`, so the reachable misroute is a SELECT carrying the token in a literal:

```sql
SELECT name FROM lit_c2 WHERE name = 'ndb_chunk_text(x)'
```

That statement routed into `chunk_text::execute_chunk_text` before this PR. That is the case to pin.

Both are one-line changes. Apply the same red-then-green check you ran on the TOPK test — if a test does not fail against the reverted arm, it is not a regression pin.

Also worth renaming: the test says UPSERT and uses `INSERT`. For WEIGHTED_PICK it makes no difference (the arm never had a statement-type guard), but the name should say what it does.

## Comment hygiene — lines 5-10 and 16

The module doc narrates the pre-fix state: "The router **misrouted** statements … which **fails with** 'TOPK requires 2 arguments'. The arms **are now** anchored". A regression test does not get an exemption from the standing rule — the invariant states perfectly well in the present tense, about the code as it stands:

```rust
//! Prefix anchoring for the router's function arms.
//!
//! `TOPK`, `WEIGHTED_PICK` and `NDB_CHUNK_TEXT` are recognized only at the
//! statement prefix. A statement carrying one of those tokens anywhere else —
//! a doc-object UPSERT value, a string literal, a comment — belongs to its own
//! handler and must never reach the function arm. These tests hold both
//! directions: the literal stores verbatim, the anchored form still routes.
```

Line 16 has the same tell: "the reported bug had the UPSERT misrouting into `select_topk`". Cut the clause — the two lines above it already state the requirement.

Nothing in `string_engine_ops.rs` or `dispatch.rs` has this problem; those comments read forward correctly.

Nothing else outstanding. TOPK is pinned, the trim is right, `fmt` is clean, the new file is registered in `cases/mod.rs` and sits well under the test file limit.

EnRaiha:
All three points addressed in `98e1f25a4`:

## 1. Vacuous second test — now pins, red-proven

Split into two tests with the correct literal shapes:

- `insert_literal_weighted_pick_stores_verbatim` — stores `weighted_pick()` (call order, so `contains("WEIGHTED_PICK(")` would match it) and reads it back verbatim.
- `select_literal_does_not_misroute_ndb_chunk_text` — `SELECT name FROM lit_c3 WHERE name = 'ndb_chunk_text(x)'` returns the matching row, not chunk-text output. That is the reachable misroute for the old `starts_with("SELECT ") && contains(...)` predicate.

Red-then-green applied to both, the same as the TOPK test: reverting the WEIGHTED_PICK and NDB_CHUNK_TEXT arms to their old `contains` forms makes both fail (the WEIGHTED_PICK INSERT errors, the chunk-text SELECT returns chunks), restoring the anchors makes them pass.

## 2. Rename

The weighted_pick test is now `insert_literal_weighted_pick_stores_verbatim` — it uses INSERT, and the name says what it does.

## 3. Comment hygiene

Module doc rewritten to the present tense, per your text:

```
Prefix anchoring for the router's function arms.

TOPK, WEIGHTED_PICK and NDB_CHUNK_TEXT are recognized only at the
statement prefix. A statement carrying one of those tokens anywhere else —
a doc-object UPSERT value, a string literal, a comment — belongs to its own
handler and must never reach the function arm.
```

The `reported bug had …` clause is cut, and the two inline pre-fix narrations in the test bodies are removed. Swept the file — no issue numbers, no links, no past-tense state descriptions.

## Validation

- Wire `router_misroute_literals`: 4/4
- Wire `sorted_index` + `query_function_authorization` + `router_misroute_literals`: 33/33

PR #270, opened 2026-09-04
fix(sql,vector): reject negative LIMIT, bound ANALYZE name, unify oversample

### LINKED ISSUE #272: LIMIT -1 is silently treated as unbounded instead of rejected
origin/main @ 7205d9fd9

### Deployment mode
Origin — single node (local)

### Engine(s) involved
Not engine-specific / unsure

### Summary
A negative LIMIT literal — `LIMIT -1`, `LIMIT '-1'` (untyped parameter), or `OFFSET -2` — is silently dropped and the query runs unbounded. PostgreSQL rejects these with SQLSTATE 2201W (invalid_limit_value). The planner's literal extraction returns `Option<usize>`, and `-1` cannot parse as `usize`, so the bound collapses to `None` = no limit.

### Steps to reproduce
From a fresh data directory, with a collection that holds more rows than you expect:
```sql
CREATE COLLECTION t (id TEXT PRIMARY KEY) WITH (engine='document_strict');
-- insert 5 rows
INSERT INTO t (id) VALUES ('r1'), ('r2'), ('r3'), ('r4'), ('r5');
SELECT * FROM t LIMIT -1;
-- returns all 5 rows (expected: error)
SELECT count(*) FROM t LIMIT 0;
-- returns 1 row with count 5 (correct)
```

### Expected behavior
`LIMIT -1` returns an error (`invalid limit value` family), as PostgreSQL does with 2201W. No rows come back.

### Actual behavior
The query succeeds and returns the whole collection — the negative bound is dropped entirely. A careless negative bound on a huge collection scans everything.

### Proposed severity

EnRaiha:
## Summary

Three planner/parsing correctness fixes, each with a regression test:

| Finding | Fix |
|---------|-----|
| `LIMIT -1` silently unbounded (#272) | Fallible `expr_as_nonnegative_usize`; `limit_offset()` returns `Result`; statically-negative bounds (all literal forms: `-1`, `'-1'`, `- '-1'`) rejected as `InvalidLimitValue`, carried to pgwire as SQLSTATE `2201W`; non-literal operands (`LIMIT -$1`) keep the documented unbounded semantics |
| `ANALYZE users(id)` absorbs `(id)` (#273) | Keyword-boundary match on the first token (whitespace-safe), parenthesised payload stripped |
| Vector cost planner oversample 3 vs BBQ default 4 (#275) | Single source of truth: `rerank::codecs::bbq::DEFAULT_OVERSAMPLE` |

## How to test

```bash
cargo test -p nodedb-sql --lib s1_negative_limit_rejected_not_unbounded
cargo test -p nodedb-sql --lib s2_analyze
cargo test -p nodedb-vector --lib v3_oversample_constant_matches_bbq
```

Manual:
- `SELECT * FROM t LIMIT -1` → error `invalid limit value` (2201W), not unbounded
- `SELECT * FROM t LIMIT 10` → unchanged
- `ANALYZE users(id)` → targets collection `users`
- `ANALYZE\tusers` (tab separator) → targets `users`

## Regression
- `cargo test -p nodedb-sql --lib`: 863 passed
- `cargo test -p nodedb-vector --lib v3_`: passed
- `cargo test -p nodedb --lib error_map`: 29 passed
- Lateral-planner LIMIT path intentionally unchanged (documented permissive)

Fixes #272, #273, #275.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1299,9 +1299,9 @@ checksum = "f079e83a288787bcd14a6aea84cee5c87a67c5a3e660c30f557a3d24761b3527"
 
 [[package]]
 name = "chacha20"
-version = "0.10.1"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+checksum = "65c35e4b699c7e15ccbe7ee35c005e4fc0a278d22238a2857e6ce2dadeda1b06"
 dependencies = [
  "cfg-if",
  "cpufeatures 0.3.0",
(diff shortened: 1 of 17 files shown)


EnRaiha:
### Supply-chain note — zerompk git source (deny.toml allow-git)

Reviewer flagged the `allow-git` shift from `farhan-syah/zerompk` → `nuskey8/zerompk`. Recording the decision explicitly for the maintainer:

- `Cargo.toml` has pinned `nuskey8/zerompk` @ `75d527f` since commit `156ed7b02` (*build: move zerompk pin upstream, enable redb experimental cursor API*, authored by Farhan Syah) — nuskey8 is the **upstream** repo; `farhan-syah/zerompk` was the earlier *personal fork* (`f0f68d343`), not the other way around.
- The switch to upstream main carries the `inline(always)` codec fix that keeps large plan enums from taking minutes to compile; the fork does not.
- This PR only fixes the stale deny config so the gate matches what Cargo.toml already ships. No dependency code changes.
- `deny.toml` comment kept with review-by discipline; drop the git pin entirely once the fix lands in a zerompk crates.io release.

PR #299, opened 2026-09-06
fix(sql): raise 42703 for unknown column references instead of folding to NULL

### LINKED ISSUE #292: Unknown column references resolve to NULL instead of raising 42703 — WHERE/ORDER BY/UPDATE silently misbehave, including on fixed-schema engines

origin/main @ 288615500

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Document (schemaless), Document (strict), Key-Value

### Summary

A column name that does not exist in the collection resolves to NULL everywhere instead of raising `42703` (undefined_column). Projection returns NULL-filled rows, `WHERE` treats the column as NULL (so `= <value>` silently matches nothing and `IS NULL` matches everything), `ORDER BY` silently no-ops, and `UPDATE ... WHERE <unknown> = ...` silently updates zero rows. This affects the strict and kv engines too, where the schema is fixed and a missing field cannot be schemaless-by-design. Every typo'd identifier in an application query returns plausible-but-wrong results with no error — the same silent-folding class as the undefined-function and division-by-zero paths that now raise, but on the column-resolution side, which still folds.

### Steps to reproduce

```sql
CREATE COLLECTION probe (id INT PRIMARY KEY, x INT);
INSERT INTO probe (id, x) VALUES (1, 1);
INSERT INTO probe (id, x) VALUES (2, 2);

SELECT nonexistent_col FROM probe;
-- 2 rows, all NULL          <- expected ERROR 42703

SELECT count(*) FROM probe WHERE nonexistent_col = 1;
-- 0                         <- expected ERROR 42703; silently matches nothing

SELECT count(*) FROM probe WHERE nonexistent_col IS NULL;
-- 2                         <- silently matches everything

SELECT x FROM probe ORDER BY nonexistent_col;
-- rows in storage order     <- sort key silently ignored

UPDATE probe SET x = 99 WHERE nonexistent_col = 1;
-- UPDATE 0                  <- write silently no-ops

-- Fixed-schema engines fold identically:
CREATE COLLECTION probe_strict (a INT4 PRIMARY KEY, b INT8) WITH (engine = 'document_strict');
INSERT INTO probe_strict (a, b) VALUES (1, 2);
SELECT nonexistent_col FROM probe_strict;   -- 1 row, NULL  <- expected ERROR 42703

CREATE COLLECTION probe_kv (k TEXT PRIMARY KEY, v TEXT) WITH (engine = 'kv');
INSERT INTO probe_kv (k, v) VALUES ('a', '1');
SELECT nonexistent_col FROM probe_kv;       -- 1 row, NULL  <- expected ERROR 42703
```

### Expected behavior

An identifier that names no column of any collection in scope raises `42703` (undefined_column) at plan time, in every clause (projection, WHERE, GROUP BY, ORDER BY, UPDATE SET/WHERE), matching PostgreSQL. For the document (schemaless) engine, a declared-schema collection knows its column set at plan time just as strict does; if resolving unknown identifiers to NULL is intended for genuinely schemaless reads, that intent cannot extend to `document_strict`/`kv`, whose schemas are closed.

### Actual behavior

The unknown identifier is planned as a field reference and evaluates to NULL per row on every engine tested, including `document_strict` and `kv`. Predicates, sort keys, and write predicates built on the typo silently degrade: reads return wrong row sets, `UPDATE`/`DELETE` silently touch zero rows, and no error surfaces anywhere.

- [x] Acknowledged/committed data was lost, corrupted, or silently wrong — *query results silently wrong; stored data intact*
- [x] Core functionality is broken with no acceptable workaround

### Proposed severity

### Reproducibility

Always — every attempt

Unknown — likely never raised; the plan-time existence gate added for functions does not cover column references.

### Environment & logs

Linux x86_64, release build from a fresh data directory, trust mode, pgwire via psql 16. No relevant server log lines — no error is produced anywhere on these paths.

### Before submitting

EnRaiha:
fix(sql): raise 42703 for unknown column references instead of folding to NULL

Fixes #292

## Problem

A column reference that names nothing in scope planned as a field lookup and evaluated to NULL on every row. On collections whose row shape is known at plan time that is silent misbehavior: projection returned NULL-filled rows, WHERE matched nothing (IS NULL matched everything), ORDER BY no-oped, and UPDATE/DELETE predicates touched zero rows — no error anywhere. This is the same folding class as the undefined-function and division-by-zero paths that now raise (see the #226 fix), on the column-resolution side, which still folded.

## Fix

Closed-schema collections validate every column reference at plan time and raise SQLSTATE 42703 (undefined_column):

- `document_strict`, `kv`, `columnar`, `timeseries` and `spatial` are closed by construction.
- `document` (schemaless) is closed when the collection declares columns. The open/closed line is the declared column count, not a name: the catalog adapter always prepends the synthesized primary-key column (the declared PK name or the built-in `id`), so a one-column list means nothing was declared and dynamic fields keep resolving — the NULL fold stays the documented behavior there. Declared fields close the set even though schemaless ColumnInfo carries `raw_type: None`.
- Schemaless writes stay open: INSERT column lists may still carry undeclared dynamic fields.

The gate covers the single-table SELECT path (projection, WHERE, GROUP BY, HAVING, ORDER BY) and the UPDATE/DELETE target path (SET targets and predicates). Qualified references validate against the collection they name.

The error is typed end to end: `SqlError::UnknownColumn` maps to a new `crate::Error::UndefinedColumn`, to the public `NodeDbError` code 1206 (msgpack tag 79), and to SQLSTATE 42703 on pgwire and the native protocol — mirroring the `undefined_function` path added for the same folding class.

## Out of scope

Pre-existing folds, tracked separately: JOIN/lateral/derived paths, aggregate ORDER BY, UPDATE ... FROM predicates, INSERT ... SELECT.

## Verified

| Query (declared doc / strict / kv / columnar) | Before | After |
|---|---|---|
| `SELECT ghost` | N rows NULL | 42703 |
| `WHERE ghost = 1` / `IS NULL` | 0 / all rows | 42703 |
| `ORDER BY ghost` | storage order | 42703 |
| `UPDATE/DELETE ... WHERE ghost` | 0 rows | 42703 |
| auto-id / renamed-PK-only schemaless `SELECT dyn_field` | fold / resolves | unchanged |

- Unit: `cargo nextest run -p nodedb-sql --lib` — 875 passed (11 validator tests: every clause, closed/open/renamed-PK, function-arg walk, GROUP BY alias, sort-key ordinals)
- Unit: `cargo nextest run -p nodedb-types --lib` — 688 passed
- Wire (new `undefined_column_42703`): 4 passed — all six clause shapes on the default-declared engine; strict/kv/columnar; open fold incl. renamed-PK carve-out
- Regression sentinels: `engine_surface_document_schemaless`, `sql_json_path_ops`, `ddl_numeric_width_schemaless`, `sql_typeguard_defaults` — 49 passed
- `cargo fmt --all -- --check` clean; clippy `-D warnings` clean on nodedb-sql, nodedb-types, nodedb

## How to test

```
cargo nextest run -p nodedb-sql --lib
cargo nextest run -p nodedb-types --lib
cargo nextest run -p nodedb --test wire -E 'test(undefined_column_42703) or test(engine_surface_document_schemaless) or test(sql_json_path_ops) or test(ddl_numeric_width_schemaless) or test(sql_typeguard_defaults)'
```

### DIFF
--- a/nodedb-sql/src/planner/dml_update_delete.rs
+++ b/nodedb-sql/src/planner/dml_update_delete.rs
@@ -50,7 +50,21 @@ pub fn plan_update(stmt: &ast::Statement, catalog: &dyn SqlCatalog) -> Result<Ve
             name: table_name.clone(),
         })?;
 
+    // Closed-schema existence gate (42703): a WHERE or SET target naming a
+    // column the collection does not declare must not silently no-op.
+    let resolved = crate::resolver::columns::ResolvedTable {
+        name: table_name.clone(),
+        alias: None,
+        info: info.clone(),
+    };
+    if let Some(selection) = &update.selection {
+        crate::planner::select::validate_columns::validate_where(selection, &resolved)?;
+    }
+
     let mut assigns = convert_assignments(&update.assignments)?;
+    for (column, _) in &assigns {
+        crate::planner::select::validate_columns::validate_write_column(column, &resolved)?;
+    }
     // Re-type each literal assignment to its declared column type before the
     // range check reads it — the same order, and for the same reason, as the
     // INSERT path's `coerce_and_check_rows`. Engines with a typed write path
@@ -357,6 +371,16 @@ pub fn plan_delete(stmt: &ast::Statement, catalog: &dyn SqlCatalog) -> Result<Ve
         .ok_or_else(|| SqlError::UnknownTable {
             name: table_name.clone(),
         })?;
+
+    // Closed-schema existence gate (42703) for the DELETE predicate.
+    let resolved = crate::resolver::columns::ResolvedTable {
+        name: table_name.clone(),
+        alias: None,
+        info: info.clone(),
+    };
+    if let Some(selection) = &delete.selection {
+        crate::planner::select::validate_columns::validate_where(selection, &resolved)?;
+    }
 
     let filters = match &delete.selection {
         Some(expr) => super::super::select::convert_where_to_filters(expr)?,
--- a/nodedb-sql/src/planner/select/mod.rs
+++ b/nodedb-sql/src/planner/select/mod.rs
@@ -15,6 +15,7 @@ mod order_by;
 mod post_process;
 mod query_tail;
 mod select_stmt;
+pub(crate) mod validate_columns;
 mod where_search;
 
 pub use entry::plan_query;
--- a/nodedb-sql/src/planner/select/select_stmt.rs
+++ b/nodedb-sql/src/planner/select/select_stmt.rs
@@ -20,6 +20,8 @@ use crate::resolver::columns::TableScope;
 use crate::temporal::TemporalScope;
 use crate::types::*;
 
+use super::validate_columns;
+
 /// Plan a single SELECT statement (no UNION, no CTE wrapper).
 ///
 /// `tail` carries the enclosing query's ORDER BY / LIMIT so the base scan can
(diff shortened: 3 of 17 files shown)


farhan-syah:
Superseded by #304, merged as `66d0a225c`. #292 is closed.

#304 threads a `ColumnScope` through `convert_expr`, the single point every clause funnels an identifier through. The gate therefore reaches JOIN, LATERAL, derived tables, aggregate arguments, `UPDATE ... FROM` predicates, and `INSERT ... SELECT`.

The two differ on one point, worth recording for anyone reading back. This PR closes a schemaless collection once it declares columns. #304 keeps `document_schemaless` open whatever its declared column count. `docs/documents.md:43` states undeclared fields pass freely on write, so closing the read side makes written data unreadable.

#304 was written independently and did not draw on this PR.

PR #302, opened 2026-09-06
fix(sql): raise 23502 for NULL primary keys instead of committing them

### LINKED ISSUE #293: Document and kv engines accept NULL primary keys; multiple NULL-pk rows coexist and IS NULL answers differ between scan and aggregate paths

origin/main @ 288615500

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Document (schemaless), Key-Value

### Summary

The document (schemaless) and kv engines accept NULL into a declared PRIMARY KEY — both as an explicit `NULL` value and by omitting the column — and multiple NULL-pk rows coexist in one collection. Uniqueness enforcement works for real values (a duplicate is correctly rejected) but NULL bypasses it entirely, along with the implied NOT NULL. On top of that, the stored NULL-pk rows are internally inconsistent: `SELECT ... WHERE id IS NULL` returns both rows while `SELECT count(*) ... WHERE id IS NULL` returns 1 — the same predicate gives different answers on the scan and aggregate paths, suggesting one row stores an explicit null and the other has the field absent, and the two paths treat absence differently. `document_strict` correctly rejects both inserts, so the hole is confined to the engines that route through the schemaless write path.

### Steps to reproduce

```sql
CREATE COLLECTION pk2 (id INT PRIMARY KEY, v TEXT);

INSERT INTO pk2 (id, v) VALUES (NULL, 'explicit-null');   -- INSERT 0 1   <- expected ERROR 23502
INSERT INTO pk2 (v) VALUES ('omitted-pk');                -- INSERT 0 1   <- expected ERROR 23502

SELECT id, v FROM pk2;
--  id |       v
-- ----+---------------
--     | explicit-null
--     | omitted-pk
-- (2 rows)                  <- two rows with NULL primary key coexist

SELECT count(*) FROM pk2 WHERE id IS NULL;   -- 1     <- WRONG, 2 rows match
SELECT id, v FROM pk2 WHERE id IS NULL;      -- 2 rows <- same predicate, different answer

-- Uniqueness machinery does exist for real values:
INSERT INTO pk2 (id, v) VALUES (1, 'ok');    -- INSERT 0 1
INSERT INTO pk2 (id, v) VALUES (1, 'dup');   -- ERROR: duplicate key ... primary-key uniqueness (correct)

-- kv accepts it too:
CREATE COLLECTION pk4 (k TEXT PRIMARY KEY, v TEXT) WITH (engine = 'kv');
INSERT INTO pk4 (v) VALUES ('kv-omitted');   -- OK           <- expected ERROR 23502

-- document_strict rejects both (correct), though via a serialization error rather than 23502:
CREATE COLLECTION pk3 (id INT PRIMARY KEY, v TEXT) WITH (engine = 'document_strict');
INSERT INTO pk3 (id, v) VALUES (NULL, 'strict-null');
-- ERROR: serialization error (binary_tuple): bad request: column 'id' is NOT NULL but no value provided
```

### Expected behavior

PRIMARY KEY implies NOT NULL on every engine: both inserts raise `23502` (not_null_violation) and no NULL-pk row is ever stored. Predicates over the pk column return the same row set on every execution path. (Secondary: the strict engine's rejection would ideally surface as a clean `23502` constraint error rather than a `serialization error (binary_tuple)` message, so drivers classify it correctly.)

### Actual behavior

Document and kv engines commit NULL-pk rows — several per collection — so the primary key is neither unique nor non-null for these rows. The rows then behave inconsistently: the aggregate path counts 1 while the scan path returns 2 for the identical `IS NULL` predicate, so applications cannot even reliably find the damage. `DELETE ... WHERE id IS NULL` removes 2 rows, agreeing with the scan path and disagreeing with `count(*)`.

- [x] Acknowledged/committed data was lost, corrupted, or silently wrong — *rows violating the declared key constraint are committed and readback of them is inconsistent; other stored data intact*
- [x] Core functionality is broken with no acceptable workaround

### Proposed severity

### Reproducibility

Always — every attempt

Unknown.

### Environment & logs

Linux x86_64, release build from a fresh data directory, trust mode, pgwire via psql 16. No relevant server log lines — every accepting statement completes without warning.

### Before submitting

EnRaiha:
fix(sql): raise 23502 for NULL primary keys instead of committing them

Fixes #293

## Problem

A NULL primary key — explicit, or by omitting the column — committed on the document (schemaless) and kv engines. Uniqueness still applied to real values, so the key was neither unique nor non-null for those rows, and several NULL-keyed rows coexisted per collection. Readback was silently inconsistent: `SELECT` and `DELETE` found all NULL-keyed rows while `count(*)` found one — one row stored an explicit null, the other had the field absent, and the two paths treated absence differently.

## Fix

PRIMARY KEY implies NOT NULL at plan time on every engine with a declared key. `INSERT`, `UPSERT`, and `INSERT ... ON CONFLICT` rows are checked after type coercion, before engine dispatch:

| Input | Before | After |
|---|---|---|
| `INSERT ... (id, v) VALUES (NULL, ...)` declared doc | committed NULL-key row | `23502` |
| `INSERT ... (v) VALUES (...)` (pk omitted), declared doc | committed NULL-key row | `23502` |
| Same on kv | committed empty-key entry | `23502` |
| Same on `document_strict` | internal `serialization error (binary_tuple)` | clean `23502` |
| `UPSERT` with NULL key (doc/kv) | committed | `23502` |
| Auto-id / renamed-PK-only schemaless, key omitted | fresh identity minted | unchanged (synthetic-key exemption) |
| Duplicate real key | `23505` uniqueness | unchanged |

A row that omits the declared key or binds it to NULL used to fall through to the "fresh surrogate" identity path — silently minting a new identity for what the caller declared as the row's key. Collections whose key is synthetic legitimately mint fresh identities and are exempt: `_rowid` collections and schemaless collections created without a column list (their only catalog column is the auto PK, matching the #292 closed-schema line).

The error is typed end to end: `SqlError::NotNullViolation` → `crate::Error::NotNullViolation` → public `NodeDbError` code 1207 / msgpack tag 80 → SQLSTATE 23502 on pgwire and native, mirroring the `undefined_function` path.

## Out of scope

- Pre-existing NULL-keyed rows committed before this fix remain; repair of existing data is tracked separately.
- `INSERT ... SELECT` sources flow through a separate conversion path; a NULL key there still takes the old surrogate route (tracked separately).
- Raw kv collections (no declared PK, sentinel `key` column): a NULL/omitted key is rejected by the same gate — the key bytes are the identity there too.
- `PRIMARY KEY DEFAULT NULL` (pathological DDL) can still reach the surrogate route when the default evaluates to nothing; noted, not chased.

## How to test

```
cargo test -p nodedb-sql --lib
cargo test -p nodedb-types --lib
cargo nextest run -p nodedb --test wire -E 'test(not_null_pk_23502)'
```

## Verified

- Live matrix on fresh main + this branch: 23502 on declared-doc / document_strict / kv for explicit-NULL and omitted keys across INSERT/UPSERT; uniqueness (23505) unchanged; auto-id exemption intact.
- Engines gated are exactly the issue's set (document_strict, kv, declared-schema document). columnar/timeseries/spatial are deliberately NOT planner-gated: their identity contract and the Data Plane already reject a NULL key themselves (engine-side NOT NULL), so a planner gate would only duplicate an engine error.
- Wire `not_null_pk_23502`: 4 passed.
- Sentinels (kv defaults, schemaless surface, insert value expressions, KV/insert RETURNING): 47 passed.
- `cargo fmt --all -- --check` clean; clippy `-D warnings` clean on nodedb-sql, nodedb-types, nodedb; unit 864 + 688 passed.

### DIFF
--- a/nodedb-sql/src/planner/dml.rs
+++ b/nodedb-sql/src/planner/dml.rs
@@ -97,6 +97,74 @@ fn classify_on_conflict(ins: &ast::Insert, scope: &TableScope) -> Result<OnConfl
     }
 }
 
+/// PRIMARY KEY implies NOT NULL on every engine.
+///
+/// A row that omits the declared primary-key column or binds it to NULL
+/// must raise `23502` (not_null_violation) rather than commit. On the
+/// document path such a row used to fall through to "fresh surrogate" —
+/// silently minting a new identity for what the caller declared as the row's
+/// key, so several NULL-key rows coexisted and readback of them disagreed
+/// between scan and aggregate paths. Collections whose key is synthetic
+/// (`_rowid`, or a schemaless collection carrying only the auto-injected PK
+/// column) legitimately mint fresh identities and are exempt.
+fn enforce_pk_not_null(
+    engine: EngineType,
+    collection: &str,
+    primary_key: Option<&str>,
+    columns: &[ColumnInfo],
+    rows: &[Vec<(String, SqlValue)>],
+) -> Result<()> {
+    let Some(pk) = primary_key else {
+        return Ok(());
+    };
+    if pk == "_rowid" {
+        return Ok(());
+    }
+    // Closed to the engines #293 names: document_strict and kv reject NULL
+    // keys today, and a declared schemaless column list closes the document
+    // engine. columnar/timeseries/spatial are deliberately NOT gated: their
+    // identity contract mints a surrogate for an omitted key (SERIAL-style
+    // auto identity), so a missing key is legitimate there, not a violation.
+    // Closed to the engines #293 names: document_strict and kv reject NULL
+    // keys today; a schemaless document collection closes exactly when the
+    // user DECLARED its primary key (the auto-injected `id` is synthesized,
+    // `raw_type: None`, and omitting it legitimately mints an auto identity
+    // even when other fields are declared — upstream exercises that with
+    // `CREATE COLLECTION ... FIELDS (...)`). columnar/timeseries/spatial are
+    // deliberately NOT gated: their identity contract and the Data Plane
+    // handle NULL keys themselves.
+    let closed = match engine {
+        EngineType::DocumentSchemaless => columns
+            .iter()
+            .any(|c| c.is_primary_key && c.raw_type.is_some()),
+        EngineType::KeyValue | EngineType::DocumentStrict => true,
+        EngineType::Columnar | EngineType::Timeseries | EngineType::Spatial | EngineType::Array => {
+            false
+        }
+    };
+    if !closed {
+        return Ok(());
+    }
+    let pk_has_default = columns.iter().any(|c| c.name == pk && c.default.is_some());
+    for row in rows {
+        let cell = row.iter().find(|(name, _)| name == pk).map(|(_, v)| v);
+        let missing = cell.is_none();
+        let null = matches!(cell, Some(SqlValue::Null));
+        // A declared DEFAULT materializes at conversion; only absence is
+        // exempt, an explicit NULL still violates the constraint.
+        if missing && pk_has_default {
+            continue;
+        }
+        if missing || null {
+            return Err(SqlError::NotNullViolation {
+                table: collection.to_string(),
+                column: pk.to_string(),
+            });
+        }
+    }
+    Ok(())
+}
+
 /// Plan an INSERT statement.
 pub fn plan_insert(ins: &ast::Insert, catalog: &dyn SqlCatalog) -> Result<Vec<SqlPlan>> {
     let table_name = match &ins.table {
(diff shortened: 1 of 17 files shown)


nodedb-sql/src/planner/dml.rs
```diff
@@ -97,6 +97,74 @@ fn classify_on_conflict(ins: &ast::Insert, scope: &TableScope) -> Result<OnConfl
...
+    };
+    if pk == "_rowid" {
+        return Ok(());
+    }
+    // Closed to the engines #293 names: document_strict and kv reject NULL
+    // keys today, and a declared schemaless column list closes the document
+    // engine. columnar/timeseries/spatial are deliberately NOT gated: their
+    // identity contract mints a surrogate for an omitted key (SERIAL-style
+    // auto identity), so a missing key is legitimate there, not a violation.
+    // Closed to the engines #293 names: document_strict and kv reject NULL
```
farhan-syah: Issue number leak. And make all comments more concise,not verbose

farhan-syah:
Superseded by #310

Issue #298, opened 2026-09-06
Timeseries TIME_KEY declared TIMESTAMP returns OID 25 text carrying raw epoch millis instead of a timestamp


emanzx:
### Version / build tested against

origin/main @ 288615500

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Timeseries

### Summary

A timeseries `TIME_KEY` column declared `TIMESTAMP` comes back over pgwire as OID 25 (TEXT) carrying the raw epoch-milliseconds integer as a string (`'1788692400000'`), instead of OID 1114 with a timestamp rendering. Every consumer downstream misreads it: psql displays a bare number, drivers hand applications a string, ORMs fail to map the column to a datetime type, and BI tools cannot treat the time axis as time. The declared DDL type and the wire type disagree — the same fidelity contract that RowDescription preserves for integer widths on other engines is dropped on the timeseries time key.

### Steps to reproduce

```sql
CREATE COLLECTION tsp (ts TIMESTAMP TIME_KEY, val DOUBLE PRECISION) WITH (engine = 'timeseries');
INSERT INTO tsp (ts, val) VALUES ('2026-09-06T10:00:00Z', 1.5);

SELECT ts, val FROM tsp;
--       ts       | val
-- ---------------+-----
--  1788688800000 | 1.5      <- epoch millis as text; expected 2026-09-06 10:00:00
```

Driver-level (psycopg2) RowDescription:

```
cur.execute('SELECT ts, val FROM tsp')
cur.description  -> [('ts', 25), ('val', 701)]     # ts = OID 25 (TEXT); expected 1114 (TIMESTAMP)
cur.fetchone()   -> ('1788692400000', 2.5)         # string of millis
```

### Expected behavior

The wire type follows the declared type: OID 1114 (TIMESTAMP) in RowDescription and a PostgreSQL-format timestamp in the text protocol (`2026-09-06 10:00:00`), so psql, drivers, and ORMs read the time key as a timestamp. Ingest already accepts the RFC3339 timestamp literal, so the asymmetry is only on the read side.

### Actual behavior

RowDescription declares TEXT and the value is the internal epoch-millis representation serialized as a string. Round-tripping a value the server itself accepted (`INSERT ... SELECT`-style workflows, dump/reload) yields a format the insert path does not document as canonical, and applications must special-case the column.

### What actually happened? (check all that are true)

- [ ] Acknowledged/committed data was lost, corrupted, or silently wrong
[...]

Issue #295, opened 2026-09-06
Expression errors fold silently over constant derived tables — 22012 raised on collection scans but not on FROM (SELECT ...)


emanzx:
### Version / build tested against

origin/main @ 288615500

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Not engine-specific / unsure

### Summary

Expression errors still fold silently when the query's FROM source is a constant derived table (`FROM (SELECT 1 AS x) s`). Division by zero in a projection, aggregate argument, GROUP BY key, or window PARTITION BY over such a derived table returns NULL rows or an empty result set with success, instead of raising `22012`. The identical expressions over a real collection raise correctly — the earlier conversions of the aggregate/GROUP BY/window paths hold there — so the remaining hole is specific to the plan shape produced for constant subquery sources, which appears to pre-evaluate the expression on a path that still has the silent fallback. In the GROUP BY case a projected column is additionally dropped from the result set entirely.

### Steps to reproduce

```sql
-- Constant scalar raises (correct):
SELECT 1/0;                                      -- ERROR: division by zero

-- Same expressions over a real collection raise (correct):
CREATE COLLECTION divp (id INT PRIMARY KEY, x INT);
INSERT INTO divp (id, x) VALUES (1, 1);
SELECT sum(x/0) FROM divp;                       -- ERROR: division by zero
SELECT x, count(*) FROM divp GROUP BY x/0;       -- ERROR: division by zero
SELECT sum(x) OVER (PARTITION BY x/0) FROM divp; -- ERROR: division by zero

-- Over a constant derived table, every one of them folds:
SELECT x/0 FROM (SELECT 1 AS x) s;
--  x / 0
-- -------
--
-- (1 row)          <- NULL, success; expected ERROR 22012

SELECT sum(x/0) FROM (SELECT 1 AS x) s;
-- 1 row, NULL      <- expected ERROR 22012

SELECT x, count(*) FROM (SELECT 1 AS x) s GROUP BY x/0;
-- (0 rows), and only the count(*) column is present in the result description
--                  <- expected ERROR 22012; also drops the projected x column

SELECT sum(x) OVER (PARTITION BY x/0) FROM (SELECT 1 AS x) s;
-- 1 row, NULL      <- expected ERROR 22012

-- The derived table itself is not the trigger — one that scans a collection raises:
SELECT sum(x/0) FROM (SELECT x FROM divp) s;     -- ERROR: division by zero
```

### Expected behavior

[...]

Issue #294, opened 2026-09-06
CREATE SEQUENCE succeeds but nextval/currval raise 42883; DEFAULT nextval(...) passes DDL then silently writes NULL


emanzx:
### Version / build tested against

origin/main @ 288615500

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Not engine-specific / unsure

### Summary

`CREATE SEQUENCE` succeeds, but the sequence cannot be used: `nextval('name')` (and `currval`/`setval`) raise `42883` "function nextval(...) does not exist" because the resolver's plan-time function-existence gate consults a registry that does not contain the sequence accessors. Worse, `DEFAULT nextval('name')` is accepted at DDL time on a column — including a PRIMARY KEY — and then silently evaluates to NULL on every insert, so the table fills with NULL keys and no error ever surfaces. The DDL path and the expression path disagree about whether `nextval` exists, and the runtime DEFAULT path still has the silent NULL fallback that the plan-time gate was built to prevent.

### Steps to reproduce

```sql
CREATE SEQUENCE seqprobe;          -- CREATE SEQUENCE (accepted)

SELECT nextval('seqprobe');
-- ERROR:  function nextval(...) does not exist    <- the object just created is unreachable

SELECT currval('seqprobe');
-- ERROR:  function currval(...) does not exist

-- The DDL path happily accepts the same function the query path rejects:
CREATE COLLECTION seqt (id BIGINT DEFAULT nextval('seqprobe') PRIMARY KEY, v TEXT);
-- CREATE COLLECTION

INSERT INTO seqt (v) VALUES ('one');   -- INSERT 0 1
INSERT INTO seqt (v) VALUES ('two');   -- INSERT 0 1

SELECT id, v FROM seqt;
--  id |  v
-- ----+-----
--     | one
--     | two
-- (2 rows)        <- DEFAULT silently produced NULL into a PRIMARY KEY, twice
```

### Expected behavior

Either the sequence accessors work end to end — `nextval`/`currval`/`setval` registered as callable functions, resolved against the sequence catalog, `SELECT nextval('seqprobe')` returns 1 then 2 — or `CREATE SEQUENCE` and `DEFAULT nextval(...)` are rejected up front as unsupported. In no case should a DDL-accepted DEFAULT expression evaluate to silent NULL at insert time; if the expression cannot be evaluated, the insert (or better, the DDL) should fail loudly.

### Actual behavior

[...]

PR #303, opened 2026-09-07
sequence accessors: DEFAULTs, loud classification, constant-context eval, window-over-derived, kv INSERT..SELECT sources

EnRaiha:
## Scope

Sequence accessors (`nextval` / `currval` / `setval`) end to end:

1. **Per-row DEFAULTs on every engine** — `DEFAULT nextval('s')` advances the control-plane registry once per inserted row on kv, columnar, document and strict collections; malformed or `currval`/`setval` DEFAULTs are rejected loudly instead of silently NULLing.
2. **Loud classification everywhere else** — an accessor in any row-scope context (SELECT list, WHERE, ORDER BY, GROUP BY, HAVING, UPDATE SET, JOIN ON, INSERT..SELECT) raises SQLSTATE 0A000 with a self-describing message; the registry entries, scalar-evaluator guard, exhaustive const-fold classification and end-to-end error plumbing (error code, msgpack tag, envelope + data-plane wire, pgwire/gateway maps) carry the typed code to the client.
3. **Real evaluation in constant contexts** — FROM-less `SELECT nextval('s')` and explicit `VALUES` cells evaluate through the registry at plan time, in expression order; `setval` sets and returns; a missing sequence raises a plan error naming it. Folded accessor plans are never admitted to the physical-plan cache, and `EXPLAIN` does not advance the registry.
4. **kv SELECT projections evaluate** — kv scans carry the SELECT projection list and computed columns like the other engines, so scalar expressions evaluate per row (`SELECT 1 + 1 FROM kv` no longer returns an empty column); plain projections stay full-row so clone-source merges keep the keys tombstone suppression needs.
5. **Typed error lanes** — executor paths that collapsed evaluation errors to division-by-zero or internal error now discriminate variants, so division keeps 22012 and accessors surface 0A000 through every scan/aggregate/join/routing path.
6. **Runtime headroom** — tokio worker stack raised to 16 MiB; deep DDL planning no longer dies with a silent SIGSEGV.

7. **Typed unknown-sequence DEFAULTs** — a DEFAULT naming a missing sequence
   raises SQLSTATE 42704 (undefined_object), matching PostgreSQL; other
   registry failures keep the plan-error class.

8. **Window functions over derived tables** — `SUM(n) OVER (...)`,
   `row_number() OVER (...)`, and every other window verb over a
   `FROM (SELECT ...)` body evaluate per row through the materialized-row
   scan instead of silently answering NULL (issue #295 window form). Window
   specs survive CTE inlining for Scan-bodied derived tables, and subquery
   fold + validation walk window expressions so catalog casts inside
   PARTITION BY / ORDER BY fold and validate.

9. **kv INSERT..SELECT sources** — `INSERT INTO <target> SELECT ... FROM
   <kv source>` copies rows through the KV materialize scan instead of
   silently reporting `INSERT 0 0` (issue #311). The copy orchestrator
   routes the source scan by engine, in-transaction copies read their own
   writes, and columnar sources refuse loudly until a columnar copy scan
   exists.

Rebased on current main (includes the primary-key NOT NULL / row-identity
change #310): DEFAULT expansion runs before identity resolution, so a
sequence-defaulted primary key is filled first and the NOT NULL
enforcement only rejects keys that are still NULL after defaults.

## Tests

Wire suites: per-engine DEFAULT, expression-context (FROM-less, SELECT
list, WHERE, ORDER BY, VALUES, INSERT..SELECT, derived-table constants),
constant-context (advance, currval, setval, EXPLAIN, missing-sequence
errors), kv projection, a 60-case engine x accessor x context matrix
asserting loud 0A000, window-over-derived positive-value (SUM,
row_number) and division-by-zero, and kv INSERT..SELECT (autocommit and
in-transaction staged). Verified against the #310 suites on the merged
tree: sql_primary_key_nullability, sql_null_predicate_parity,
write_verdict_sqlstate, plus the division-by-zero and clone regressions.

## Known boundaries

Row-scope sequence accessors stay loud by design until per-row evaluation lands (future work). The remaining #295 forms —
VALUES-derived and UNION-derived bodies plus the outer-expression case —
are tracked separately. index-DDL stack overflow (#312) is fixed here.


nodedb-sql/src/visitor/plan_visitor/trait_def.rs
```diff
@@ -73,13 +73,16 @@ pub trait PlanVisitor {
     fn insert(&mut self, args: InsertVisitArgs<'_>) -> Result<Self::Output, Self::Error>;
 
     /// Handle [`SqlPlan::KvInsert`].
+    #[allow(clippy::too_many_arguments)]
```
farhan-syah: Use struct

nodedb/src/control/planner/sql_plan_convert/dml/kv_and_vector.rs
```diff
@@ -14,12 +14,15 @@ use super::super::value::{
 use super::insert::assign_for_pk;
 use nodedb_physical::physical_task::{PhysicalTask, PostSetOp};
 
+#[allow(clippy::too_many_arguments)]
```
farhan-syah: use struct

nodedb/tests/wire/cases/sequence_expression_contexts.rs
```diff
@@ -0,0 +1,170 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+//! Sequence accessors in SQL *expression* contexts must fail loudly with
+//! SQLSTATE 0A000 (`feature_not_supported`) — never silently NULL — while
+//! `DEFAULT nextval(...)` keeps working end-to-end.
+//!
+//! Grand plan (issue #294) matrix row F: accessors are stateful and
```
farhan-syah: Leak

nodedb/tests/wire/cases/sequence_default_all_engines.rs
```diff
@@ -0,0 +1,230 @@
...
+        err.contains("42704"),
+        "unknown-sequence DEFAULT must raise 42704 (undefined_object): {err}"
+    );
+    assert!(
+        err.contains("no_kv_seq"),
+        "error must name the missing sequence: {err}"
+    );
+}
+
+/// The #294 repro engine: schemaless document with a declared column list.
```
farhan-syah: Leak

nodedb/tests/wire/cases/sequence_default_all_engines.rs
```diff
@@ -0,0 +1,230 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+//! Sequence-backed DEFAULT expressions across the remaining engines.
+//!
+//! Part 1 (this PR's base) wired `DEFAULT nextval('name')` on the
```
farhan-syah: Just explain what the code do, not narate history, tellign part 1, this pr, issue xxx, etc

nodedb/tests/wire/cases/kv_select_expressions.rs
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+//! kv-engine SELECT projection expression evaluation.
+//!
+//! Previously the kv scan carried neither the SELECT projection list nor
```
farhan-syah: Don't narate history, code should be present, explaining only what the code is doing, not what used to be, which isn't alive anymore

farhan-syah:
Superseded by #316, which landed the sequence-accessor and DEFAULT work on `main` directly.

#316 covers, from this PR's scope:

| Behaviour | Where it landed |
| --- | --- |
| `nextval`/`currval`/`setval` registered and resolved | `SqlCatalog` against the sequence registry |
| Sequence-backed DEFAULTs on every engine | One shared materializer, document, kv, columnar, vector-primary |
| Loud classification of an unevaluable DEFAULT | Typed `Result`, gated at `CREATE COLLECTION` with `42883` |
| Constant-context evaluation | FROM-less `SELECT`, resolved via `catalog_expr_fold` |
| Session-scoped `currval` | Per-connection state on `ConnSession` |

It also closes #298 and #313, and adds the volatility model that keeps a non-deterministic DEFAULT from freezing into a cached plan.

Two parts of this PR are not covered by #316 and stay open as their own issues:

- kv `INSERT..SELECT` expression cells — #311
- window over a derived table — #305

Closing to clear the conflicting branch, not to discard those. Thanks for the work here — the split you documented on #294 mapped the surface accurately, and the DEFAULT classification in particular matched where the defect actually was.

PR #324, opened 2026-09-14
fix: nine-issue batch — sequences, kv scans, windows, graph DSL, and one instant conversion

### LINKED ISSUE #295: Expression errors fold silently over constant derived tables — 22012 raised on collection scans but not on FROM (SELECT ...)

### LINKED ISSUE #296: Graph DSL ignores unconsumed tokens — a typo'd clause keyword silently runs the statement with default semantics

origin/main @ 288615500

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Graph (overlay)

### Summary

### Steps to reproduce

```sql

```

### Expected behavior

### Actual behavior

### Proposed severity

### Reproducibility

Always — every attempt

### Environment & logs

### Before submitting

### LINKED ISSUE #297: Multi-row INSERT emits one CommandComplete per row — drivers report rowcount 1; kv and timeseries DML answer a bare OK tag

origin/main @ 288615500

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Document (schemaless), Key-Value, Timeseries

### Summary

A single multi-row `INSERT ... VALUES (...), (...), (...)` emits one `CommandComplete` per row (`INSERT 0 1` three times) instead of one `INSERT 0 3`. Drivers that read the affected-row count get 1 regardless of how many rows were inserted — verified with psycopg2, where `cursor.rowcount` is 1 after a 3-row insert. ORMs and applications that verify write counts (optimistic-locking checks, batch-load accounting, "exactly N inserted" assertions) see phantom under-counts. Separately, the kv and timeseries engines answer DML with a bare `OK` command tag instead of `INSERT 0 n`, which drivers cannot parse for a row count at all.

### Steps to reproduce

```sql
CREATE COLLECTION multi (id INT PRIMARY KEY, v TEXT);
INSERT INTO multi (id, v) VALUES (1,'a'), (2,'b'), (3,'c');
-- psql prints:
-- INSERT 0 1
-- INSERT 0 1
-- INSERT 0 1        <- three CommandComplete messages for ONE statement; expected one INSERT 0 3

SELECT count(*) FROM multi;   -- 3 (all rows did land)

-- kv / timeseries: no row count at all
CREATE COLLECTION kvp (k TEXT PRIMARY KEY, v TEXT) WITH (engine = 'kv');
INSERT INTO kvp (k, v) VALUES ('a', '1');
-- OK                <- expected INSERT 0 1
```

Driver-level confirmation (psycopg2):

```
cur.execute("INSERT INTO multi (id, v) VALUES (10,'x'), (11,'y'), (12,'z')")
cur.rowcount        -> 1     (expected 3)
cur.statusmessage   -> 'INSERT 0 1'
```

### Expected behavior

One statement, one `CommandComplete`: `INSERT 0 3` for a 3-row insert, per the PostgreSQL protocol contract. All engines answer DML with the standard `INSERT`/`UPDATE`/`DELETE` tags so drivers can extract the count.

### Actual behavior

The simple-query path emits one `INSERT 0 1` per value tuple; the extended-query path reports rowcount 1 for any multi-row insert. kv and timeseries DML answers `OK`, which is not a PostgreSQL command tag.

### Proposed severity

### Reproducibility

Always — every attempt

Unknown.

### Environment & logs

Linux x86_64, release build from a fresh data directory, trust mode. Tested via psql 16 (simple query) and psycopg2 (extended query).

### Before submitting

### LINKED ISSUE #305: SEARCH synthetic columns are unresolvable from a derived table over a closed-schema collection

origin/main @ 66d0a225c

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Vector

### Summary

### Steps to reproduce

```sql

SELECT s.distance

-- plans and runs

CREATE COLLECTION sp_loose;

SELECT s.distance
-- 0.0
```

### Expected behavior

### Actual behavior

### Proposed severity

### Reproducibility

Always — every attempt

### Environment & logs

### Before submitting

### LINKED ISSUE #306: ARRAY vector literal loses every element on document_strict INSERT, reported as 0 elements under XX000

origin/main @ 66d0a225c

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Document (strict)

### Summary

### Steps to reproduce

```sql

-- OK
```

| Case | Result |
|---|---|

### Expected behavior

### Actual behavior

```rust
arr.iter()
    .filter_map(|v| match v {
        _ => None,
    })
    .collect()
```

### Proposed severity

### Reproducibility

Always — every attempt

### Environment & logs

### Before submitting

### LINKED ISSUE #311: INSERT..SELECT with a kv-engine source silently NULLs expression cells
## Summary

## Evidence

## Expected

Related: #303, #307, #294
## Reproduction

```sql
```

### LINKED ISSUE #314: Evaluate sequence accessors per row in SELECT-list contexts (0A000 today)
## Summary

Sequence accessors in SELECT-list over table rows raise 0A000 by design
(constant contexts — FROM-less SELECT, VALUES — evaluate for real, see
#307). Per-row evaluation is the remaining step: `SELECT nextval('s') FROM
t` should advance per output row like PostgreSQL.

## Scope

Needs a plan-level CP marker plus a post-DP stamping stage in the control
response path that preserves output-row order (single-source scans first;
joins/unions ordering documented). RETURNING per-row evaluation rides the
same seam.

Related: #294, #303, #307

## Reproduction

```sql
CREATE SEQUENCE s;
CREATE COLLECTION t (id BIGINT, v TEXT) WITH (engine = 'kv');
INSERT INTO t (id, v) VALUES (1, 'a'), (2, 'b');
SELECT id, nextval('s') FROM t;   -- 0A000 today; PostgreSQL: advances per row
```

### LINKED ISSUE #317: Timestamp unit conversion is per-path, so a join or a cross-node scan compares mismatched units
A timeseries engine stores a time value as epoch milliseconds. The pgwire renderer expects epoch microseconds for a `DdlColType::Timestamp` cell. Four independent converters bridge that gap, one per read path.

| Converter | Path |
| --- | --- |
| `handlers/timeseries/raw_scan/row_emit.rs` | Direct scan |
| `handlers/timeseries/ingest.rs` | `RETURNING` read-back |
| `handlers/join/instant_scale.rs` | Join |
| `handlers/timeseries/encode.rs` | Grouped aggregate |

Each converter knows its own path. Nothing enforces that a new read path gets one.

## Two consequences reachable today

A remote `TimeseriesScan` gathered into a `ProviderScan` arrives in microseconds, because its own handler converted it. A local side of the same join is in milliseconds. A distributed join comparing that column in `ON` compares mismatched units.

A transforming computed expression over an instant evaluates on microseconds through a join and on milliseconds through the direct scan, because the two paths convert at different points relative to expression evaluation. A passthrough expression such as `COALESCE(ts, ts)` agrees either way. `time_bucket` over a join has no coverage.

## Expected

One converter at one boundary, replacing the four. A read path that forgets to convert must fail to compile rather than return a value that is wrong by a factor of 1000.

## Notes

Scaling inside `scan_columnar` is not the answer. Join predicates evaluate against milliseconds, so converting before predicate evaluation breaks the join.

### LINKED ISSUE #318: A cross-field typeguard guard passes the DEFAULT gate at declaration and fails at every insert

## Reproduction

```sql
```

## Expected

## Notes

EnRaiha:
### Summary

Fixes nine open issues (#295, #296, #297, #305, #306, #311, #314, #317,
#318) across SQL planning and execution, the kv engine, the graph DSL,
sequences, and the timeseries read path. 26 commits over `124cc53a6`;
verified at `fdba3e378`. A full local CI-parity run over this exact tree is
green (see Validation).

- **#295** — expression errors folded silently over constant derived tables: computed projections now evaluate on the materialized-row scan, aggregates over a non-Scan body lower to a ProviderScan sub-plan, unaliased window projections get one alias derivation, and windows evaluate in the window's own ORDER BY (`SELECT x/0 FROM (SELECT 1 AS x) s` raises `22012`). Supersedes the closed PR #321.
- **#296** — the graph DSL cursor claimed its tokens twice, so a mistyped clause key parsed as a value; the cursor now claims the tokens it reads and reports the mistyped key.
- **#297** — one `INSERT 0 n` per multi-row INSERT: plain document rows batch into one `BatchInsert` page; kv insert and put report an affected count and answer `INSERT 0 n` instead of a bare `OK`.
- **#305** — `s.distance` over a SEARCH derived table on a closed schema: derived-relation inference declares the synthetic `distance` column when the query orders by `vector_distance(...)`.
- **#306** — `ARRAY[0.1, 0.2, 0.3]` into a strict `VECTOR(3)`: element coercion accepts Float/Integer/Decimal/numeric String and rejects a non-numeric element by index and type; the `binary_tuple` wrappers preserve `BadRequest`, so a client value error is never XX000.
- **#311** — `INSERT ... SELECT` from a kv source copied 0 rows: the source scan routes on the collection's own engine (document and kv; other engines are refused by name).
- **#314** — per-row `nextval(...)` projections returned NULL: the value is stamped on the control plane at the response boundary, once per output row, across pgwire, native, and NDJSON responses. A from-less `nextval` projection evaluates; the stamp path records the session's `currval`; a missing sequence raises `42704` (the shaper keeps its own sqlstate instead of falling to XX000).
- **#317** — every read path scaled stored milliseconds itself, so a join could compare milliseconds against microseconds: the Data Plane now carries stored milliseconds end to end and one conversion runs at the response boundary, including star reads and declared time-key columns.
- **#318** — typeguard and convert identity: handlers resolve in the session database, `CONVERT` carries the source primary key, and a guard `DEFAULT` or `VALUE` that references another column is refused at declaration.

Also fixes two silent engine defects on the same paths:

- **kv scan projections** — `SELECT upper(v) FROM kv` answered NULL: `KvOp::Scan` carries the projection list and computed columns and evaluates them per row.
- **materialized-sum pages** — a multi-row INSERT into a source whose balance target is cross-shard timed out waiting for Calvin completion: such a page is un-batched to the per-row `PointInsert` shape (which settles correctly); purely co-resident pages keep their single page.

### Behaviour changes

| Path | Before | After |
|---|---|---|
| `SELECT`, `RETURNING`, native stream, NDJSON body with `nextval(...)` | NULL | one allocated value per output row |
| Window over a derived table | NULL cell | evaluated over the materialized rows, in the window's own order |
| Timeseries read after a join | unit depended on the path | milliseconds throughout, converted once at the boundary |
| `INSERT` with more than one row | one tag per row | one `INSERT 0 n` |
| `currval(...)` after a stamped `nextval(...)` | `not yet defined` | the last allocated value |
| `CONVERT` without the source key column | minted new row identities | refused with `42601` |

### Compatibility

Pre-1.0, no migration:

- The response shaper gained a sequence-stamper parameter and every response path threads the session's sequence values. Internal to the crate.
- A `CONVERT` that omits the source primary key fails instead of minting identities.
- A multi-row INSERT into a materialized-sum source with a cross-shard target plans as per-row writes (the page shape is reserved for co-resident targets). Internal to the planner.

### Validation

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --workspace --all-targets --all-features --profile ci -- -D warnings` | clean |
| static gates (`scripts/ci/*`) | 11/11 pass |
| `cargo deny check` | advisories, bans, licenses, sources all ok (lockfile bumps `rustls` to 0.23.45 for RUSTSEC-2026-0285) |
| `cargo nextest run --workspace --all-features` (fast suite) | 16,169 passed, 0 failed (5 flaky passed on retry, 3 leaky) |
| `cargo nextest run -p nodedb-cluster-tests --all-features` | 363 passed, 4 skipped (the CI-profile exclusions) |
| `cargo check --target wasm32-unknown-unknown` trio | clean |
| fuzz targets (`scripts/ci/*`) | 6/6 clean |

Local parity reaches everything except CodeQL, release packaging, and secret
scanning, which need GitHub-hosted runners; those report as not run until the
PR CI does them.

### Note on the test environment

`nextest` needs `RUST_MIN_STACK=33554432`, the value
`.github/workflows/test.yml` sets. On the 2 MiB default, two kv surrogate
tests (`kv_incr_on_fresh_key_persists_a_real_surrogate`,
`kv_transfer_persists_two_distinct_surrogates`) abort with a
`tokio-rt-worker` stack overflow — on clean `main` as well, so that gap
predates this branch. Four cluster tests excluded by `.config/nextest.toml`
from the CI profile (CI-only Calvin-completion hangs) remain excluded here.

Closes #295, closes #296, closes #297, closes #305, closes #306, closes #311, closes #314, closes #317, closes #318


EnRaiha:
Superseded: settling locally before opening for review.

EnRaiha:
Superseded by #325, which carries the settled branch at fdba3e378.

PR #325, opened 2026-09-15
fix: nine-issue batch — sequences, kv scans, windows, graph DSL, and one instant conversion

### LINKED ISSUE #295: Expression errors fold silently over constant derived tables — 22012 raised on collection scans but not on FROM (SELECT ...)

### LINKED ISSUE #296: Graph DSL ignores unconsumed tokens — a typo'd clause keyword silently runs the statement with default semantics

### LINKED ISSUE #297: Multi-row INSERT emits one CommandComplete per row — drivers report rowcount 1; kv and timeseries DML answer a bare OK tag

### LINKED ISSUE #305: SEARCH synthetic columns are unresolvable from a derived table over a closed-schema collection

### LINKED ISSUE #306: ARRAY vector literal loses every element on document_strict INSERT, reported as 0 elements under XX000

### LINKED ISSUE #311: INSERT..SELECT with a kv-engine source silently NULLs expression cells

### LINKED ISSUE #314: Evaluate sequence accessors per row in SELECT-list contexts (0A000 today)

### LINKED ISSUE #317: Timestamp unit conversion is per-path, so a join or a cross-node scan compares mismatched units

### LINKED ISSUE #318: A cross-field typeguard guard passes the DEFAULT gate at declaration and fails at every insert

EnRaiha:
### Summary

Fixes nine open issues (#295, #296, #297, #305, #306, #311, #314, #317,
#318) across SQL planning and execution, the kv engine, the graph DSL,
sequences, and the timeseries read path. 26 commits over `124cc53a6`;
verified at `fdba3e378`. A full local CI-parity run over this exact tree is
green (see Validation).

- **#295** — expression errors folded silently over constant derived tables: computed projections now evaluate on the materialized-row scan, aggregates over a non-Scan body lower to a ProviderScan sub-plan, unaliased window projections get one alias derivation, and windows evaluate in the window's own ORDER BY (`SELECT x/0 FROM (SELECT 1 AS x) s` raises `22012`).
- **#296** — the graph DSL cursor claimed its tokens twice, so a mistyped clause key parsed as a value; the cursor now claims the tokens it reads and reports the mistyped key.
- **#297** — one `INSERT 0 n` per multi-row INSERT: plain document rows batch into one `BatchInsert` page; kv insert and put report an affected count and answer `INSERT 0 n` instead of a bare `OK`.
- **#305** — `s.distance` over a SEARCH derived table on a closed schema: derived-relation inference declares the synthetic `distance` column when the query orders by `vector_distance(...)`.
- **#306** — `ARRAY[0.1, 0.2, 0.3]` into a strict `VECTOR(3)`: element coercion accepts Float/Integer/Decimal/numeric String and rejects a non-numeric element by index and type; the `binary_tuple` wrappers preserve `BadRequest`, so a client value error is never XX000.
- **#311** — `INSERT ... SELECT` from a kv source copied 0 rows: the source scan routes on the collection's own engine (document and kv; other engines are refused by name).
- **#314** — per-row `nextval(...)` projections returned NULL: the value is stamped on the control plane at the response boundary, once per output row, across pgwire, native, and NDJSON responses. A from-less `nextval` projection evaluates; the stamp path records the session's `currval`; a missing sequence raises `42704` (the shaper keeps its own sqlstate instead of falling to XX000).
- **#317** — every read path scaled stored milliseconds itself, so a join could compare milliseconds against microseconds: the Data Plane now carries stored milliseconds end to end and one conversion runs at the response boundary, including star reads and declared time-key columns.
- **#318** — typeguard and convert identity: handlers resolve in the session database, `CONVERT` carries the source primary key, and a guard `DEFAULT` or `VALUE` that references another column is refused at declaration.

Also fixes two silent engine defects on the same paths:

- **kv scan projections** — `SELECT upper(v) FROM kv` answered NULL: `KvOp::Scan` carries the projection list and computed columns and evaluates them per row.
- **materialized-sum pages** — a multi-row INSERT into a source whose balance target is cross-shard timed out waiting for Calvin completion: such a page is un-batched to the per-row `PointInsert` shape (which settles correctly); purely co-resident pages keep their single page.

### Behaviour changes

| Path | Before | After |
|---|---|---|
| `SELECT`, `RETURNING`, native stream, NDJSON body with `nextval(...)` | NULL | one allocated value per output row |
| Window over a derived table | NULL cell | evaluated over the materialized rows, in the window's own order |
| Timeseries read after a join | unit depended on the path | milliseconds throughout, converted once at the boundary |
| `INSERT` with more than one row | one tag per row | one `INSERT 0 n` |
| `currval(...)` after a stamped `nextval(...)` | `not yet defined` | the last allocated value |
| `CONVERT` without the source key column | minted new row identities | refused with `42601` |

### Compatibility

Pre-1.0, no migration:

- The response shaper gained a sequence-stamper parameter and every response path threads the session's sequence values. Internal to the crate.
- A `CONVERT` that omits the source primary key fails instead of minting identities.
- A multi-row INSERT into a materialized-sum source with a cross-shard target plans as per-row writes (the page shape is reserved for co-resident targets). Internal to the planner.

### Validation

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --workspace --all-targets --all-features --profile ci -- -D warnings` | clean |
| static gates (`scripts/ci/*`) | 11/11 pass |
| `cargo deny check` | advisories, bans, licenses, sources all ok (lockfile bumps `rustls` to 0.23.45 for RUSTSEC-2026-0285) |
| `cargo nextest run --workspace --all-features` (fast suite) | 16,169 passed, 0 failed (5 flaky passed on retry, 3 leaky) |
| `cargo nextest run -p nodedb-cluster-tests --all-features` | 363 passed, 4 skipped (the CI-profile exclusions) |
| `cargo check --target wasm32-unknown-unknown` trio | clean |
| fuzz targets (`scripts/ci/*`) | 6/6 clean |

Local parity reaches everything except CodeQL, release packaging, and secret
scanning, which need GitHub-hosted runners; those report as not run until the
PR CI does them.

### Note on the test environment

`nextest` needs `RUST_MIN_STACK=33554432`, the value
`.github/workflows/test.yml` sets. On the 2 MiB default, two kv surrogate
tests (`kv_incr_on_fresh_key_persists_a_real_surrogate`,
`kv_transfer_persists_two_distinct_surrogates`) abort with a
`tokio-rt-worker` stack overflow — on clean `main` as well, so that gap
predates this branch. Four cluster tests excluded by `.config/nextest.toml`
from the CI profile (CI-only Calvin-completion hangs) remain excluded here.

Closes #295, closes #296, closes #297, closes #305, closes #306, closes #311, closes #314, closes #317, closes #318


nodedb/src/data/executor/handlers/kv/scan.rs
```diff
@@ -155,6 +159,36 @@ impl CoreLoop {
...
+        // full-row: clone-source delegation merges rows by primary key at
+        // the control plane, then the response shape extracts columns —
+        // projecting here would strip the keys and break tombstone
+        // suppression (clone_write_suppresses_source_row).
+        if !computed_columns.is_empty() {
+            let computed_cols: Vec<crate::bridge::expr_eval::ComputedColumn> =
+                if computed_columns.is_empty() {
+                    Vec::new()
+                } else {
+                    zerompk::from_msgpack(computed_columns).unwrap_or_default()
```
habibtalib, line 178: **Inconsistent error handling + dead branch.** This block is already guarded by `if !computed_columns.is_empty()`, so the inner `if computed_columns.is_empty() { Vec::new() } else { … }` is dead — the `else` always runs. That `else` decodes with `zerompk::from_msgpack(computed_columns).unwrap_or_default()`, so malformed bytes silently yield an empty `computed_cols`, and `apply_projection_msgpack(entry, &[], projection)` then runs with no computed columns — re-opening the exact silent-NULL path this block exists to close.

The sibling `ProviderScan` path (`provider_scan.rs` ~L122) handles the identical decode by returning `ErrorCode::Internal` on failure, so the two engines diverge on the same input. In normal operation these bytes are planner-produced and decode fine, so this is a robustness/consistency issue rather than a live regression — but it's cheap to make the kv path match: drop the dead `if`, and return an error instead of `unwrap_or_default()`.

nodedb/src/data/executor/handlers/provider_scan.rs
```diff
@@ -108,7 +112,112 @@ impl CoreLoop {
...
+                            ErrorCode::Internal {
+                                detail: format!("ProviderScan: malformed window bytes: {e}"),
+                            },
+                        );
+                    }
+                };
+            if !specs.is_empty() && !rows.is_empty() {
+                let mut decoded: Vec<(String, serde_json::Value)> = Vec::with_capacity(rows.len());
+                for (i, row) in rows.iter().enumerate() {
+                    match nodedb_types::json_from_msgpack(row) {
```
habibtalib, line 187: **Window path round-trips rows through the lossy JSON codec.** This block decodes/encodes every row via `json_from_msgpack` / `json_to_msgpack`, whereas the computed-columns block just above (L135/L157) uses the lossless `value_from_msgpack` / `value_to_msgpack`. Per the repo convention (`Value` docs: JSON is intentionally lossy, MessagePack is the lossless path), a derived-table row carrying a tagged value — `Uuid`, `Bytes`, `DateTime`, `Decimal`, etc. — that is *not* touched by any window function is still downgraded to its JSON shape (e.g. `Uuid` → plain string, `Bytes` → hex string) when the whole row set is converted here. Integers survive (serde_json keeps i64), so the impact is limited to tagged types, but it's a real fidelity loss for `SELECT … OVER (…)` over a derived table containing such columns. Consider evaluating windows over the lossless `Value` codec, consistent with the computed-columns block.

nodedb-sql/src/resolver/derived.rs
```diff
@@ -86,7 +86,34 @@ fn infer_projection(
...
+    let (mut columns, open) = infer_body(catalog, &query.body)?;
+
+    // The SEARCH preprocessor rewrites `SEARCH c USING VECTOR(...)` into
+    // `SELECT * FROM c ORDER BY vector_distance(...) LIMIT k`, and the
+    // response layer appends a synthetic `distance` cell to every row of such
+    // a query. Derived-relation inference must declare that column: without
+    // it `s.distance` over a closed-schema source resolves against no
+    // relation and is refused with 42703, while the same projection over an
+    // open-schema source runs.
+    if !columns.iter().any(|c| c.name == "distance")
```
habibtalib, line 98: **Synthetic `distance` column may be phantom for non-index derived tables.** `infer_projection` appends a `distance` column for *any* derived query whose ORDER BY calls `vector_distance(...)`, but the response layer only emits a `distance` cell for plans that actually produce vector `Hit` payloads (index/SEARCH rewrite). A hand-written `SELECT s.* FROM (SELECT * FROM t ORDER BY vector_distance(col, ARRAY[…]) LIMIT k) s` over a collection whose plan is a plain scan+sort (no HNSW / brute-force order) would declare a `distance` column in the derived relation's schema that the rows never contain → `s.distance` resolves to a phantom column (schema-vs-row skew / NULL). The added test covers the SEARCH-rewritten positive case only; worth a negative test over a non-index source, and gating the synthetic column on the plan actually being a vector search.

habibtalib (commented):
Automated audit (high-effort review + manual verification against the PR tree). Findings below are posted inline. Net: one correctness/consistency fix worth making (#1), one type-fidelity concern (#2), two lower-priority items (#3 phantom column, #4 hot-path clone), and #5 — the timestamp centralization — which I traced end-to-end and found **safe** (details on that comment).

Verified clean and well-tested: graph-DSL cursor rewrite (#296), sequence-stamp plumbing (#314), sqlstate error-class fixes, cross-shard un-batching logic, typeguard database_id threading (#318), vector-element coercion (#306).

nodedb/src/control/planner/materialized_sum/resolve.rs
```diff
@@ -20,8 +20,90 @@ use super::stored::stored_row_scope;
...
+
+    for task in tasks.drain(..) {
+        let PhysicalPlan::Document(DocumentOp::BatchInsert {
+            collection,
+            documents,
+            surrogates,
+            returning,
+            rls_filters,
+            ..
+        }) = task.plan.clone()
```
habibtalib, line 54: **Hot-path clone (perf, not correctness).** `let PhysicalPlan::Document(DocumentOp::BatchInsert { .. }) = task.plan.clone() else { … }` deep-clones the entire plan — including the `documents` vector — for *every* task, even non-`BatchInsert` tasks and co-resident/no-binding batches that are pushed back unchanged. With plain multi-row INSERTs now emitted as a single `BatchInsert` page, every such insert pays a full clone of its row bodies that is immediately discarded. Matching on `&task.plan` by reference (then cloning only when a page genuinely needs un-batching) avoids copying the row payload on the common path.

nodedb/src/control/server/response_shape/compose.rs
```diff
@@ -257,6 +308,55 @@ pub fn shape_decoded_rows(
...
+/// A timeseries collection stores every instant as epoch milliseconds and the
+/// wire renders epoch microseconds. The conversion happens exactly once, here
+/// at the response boundary, after every predicate and projection — so the
+/// Data Plane, joins, and computed expressions all observe milliseconds, and a
+/// remote scan agrees with a local one.
+///
+/// Non-instant columns, SQL NULL, and cells that are not integral numbers pass
+/// through untouched. An instant that would overflow microseconds fails the
+/// response instead of wrapping.
+pub(super) fn scale_declared_instants(
```
habibtalib, line 322: **Verified safe — note on an implicit invariant.** I traced the concern that centralizing the ms→µs conversion here could double-scale engines that already store microseconds (e.g. columnar via `from_micros`). It does **not**: `scale_declared_instants` only rescales `JsonValue::Number` cells. Columnar/strict/document instants materialize as `Value::DateTime`, which serializes to an ISO-8601 **string** (`nodedb-types/src/value/json.rs:31`), so they hit the `let JsonValue::Number(n) = cell else { continue }` guard and pass through untouched. The only numeric instants reaching the shaper are timeseries (`timeseries/encode.rs:26`, `rmpv::Integer(millis)`) and document temporal columns (`document/read/audit_body.rs`, `Value::Integer(*_ms)`), both in milliseconds — consistent with this function treating a number as ms.

No change needed. The one thing worth a comment in the code: the correctness of this boundary now rests on the *implicit* invariant "every engine emits declared instants either as a `DateTime`/string or as a numeric value in **milliseconds**." A future engine that emitted a numeric Timestamp in microseconds would be silently ×1000'd here with no guard to catch it. A one-line note recording that invariant would help.

nodedb/src/data/executor/handlers/kv/scan.rs
EnRaiha: Fixed in 9fbacd7aa: the dead inner guard is gone and a malformed computed-columns blob now returns an internal error, matching the ProviderScan path.

nodedb/src/control/server/response_shape/compose.rs
EnRaiha: Comment added in 9fbacd7aa: the invariant (every engine emits declared instants as DateTime/string or as numeric milliseconds) is now recorded on scale_declared_instants.

nodedb/src/control/planner/materialized_sum/resolve.rs
EnRaiha: Fixed in 9fbacd7aa: the decision now runs on a borrow of the plan; only a page that genuinely un-batches is taken apart, so the common path clones no row bodies.

nodedb/src/data/executor/handlers/provider_scan.rs
EnRaiha: Fixed in 9fbacd7aa: window rows ride the Value codec now (evaluate_window_functions_value); only the window alias cells overlay the original row, so untouched tagged cells keep their type.

nodedb-sql/src/resolver/derived.rs
EnRaiha: Covered in 9fbacd7aa: a hand-written ORDER BY vector_distance derived table is now tested with a vector index (numeric distance cells asserted, not just the declared column) and without one (numeric cells or a refusal; silent NULL asserted against). The sort-trigger rewrite does not consult the index, so the resolver declaration matches the plan shape for Scan/Join bodies; derived.rs itself needs no change.

nodedb/tests/wire/cases/mod.rs
```diff
@@ -160,14 +162,10 @@ mod sql_bitemporal_document_visibility;
 mod sql_check_constraints;
 mod sql_collection_drop_index_cleanup;
 mod sql_conflict_policy;
-mod sql_convert_column_defs;
```
farhan-syah, line 163: **Blocker 1.** Commit `4df41f823` un-registers six wire test modules. Its message does not say so. The `.rs` files remain in `cases/` and no longer compile or run:

- `sql_convert_column_defs`
- `sql_declared_column_types`
- `sql_default_vector_primary`
- `sql_default_volatility`
- `sql_typeguard_default_gate`
- `timeseries_join_time_rendering`

`timeseries_join_time_rendering` is the direct coverage for #317. `sql_typeguard_default_gate` and `sql_convert_column_defs` cover #318. `sql_default_volatility` pins `nextval(...)` DEFAULT re-evaluation, which this PR changes.

A resubmit that touches these paths re-registers all six. A module that fails is a regression to fix, never a module to remove.

nodedb-query/src/window/eval.rs
```diff
@@ -32,6 +32,13 @@ pub fn evaluate_window_functions(
         let partitions = build_partitions(rows, &spec.partition_by)?;
 
         for partition_indices in &partitions {
+            // Evaluate the partition in the window's own ORDER BY — the
+            // query's sort, when present, orders by different keys. Without
+            // this, row_number numbered by arrival order and the rank family
+            // compared peers in the wrong sequence.
+            let ordered =
+                super::helpers::ordered_partition_indices(rows, partition_indices, &spec.order_by)?;
```
farhan-syah, line 40: **Blocker 2.** This fix lands in `evaluate_window_functions`, the JSON evaluator. After `b8d78795` no Origin path calls it. `document/read/scan.rs` and `provider_scan.rs` go through `evaluate_window_functions_on_msgpack_rows`, which calls `evaluate_window_functions_value`. That evaluator builds partitions in arrival order in `build_value_partitions`. `apply_v_row_number` numbers by that order. The scan sorts rows by the query's `ORDER BY` first. `SELECT id, row_number() OVER (ORDER BY n DESC) FROM s ORDER BY id` is still numbered by `id`.

The resubmit for #295 puts a `Value` version of `ordered_partition_indices` in `evaluate_window_functions_value`. Then remove the JSON evaluator or route it through the same helper.

nodedb/tests/wire/cases/sql_window_functions.rs
```diff
@@ -416,3 +416,126 @@ async fn window_offset_over_expression_argument_returns_previous_evaluated_value
...
+    for (i, (g, w)) in got.iter().zip(want.iter()).enumerate() {
+        assert!(
+            (g - w).abs() < 1e-9,
+            "derived SUM[{i}] = {g}, want {w}; rows = {rows:?}"
+        );
+    }
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn derived_table_window_row_number_orders_correctly() {
```
farhan-syah, line 460: Both derived-table tests order the outer query by the window key: `OVER (ORDER BY n) ... ORDER BY n`. Arrival order equals window order, so they pass with or without the ordering fix. The resubmit adds a case where the two differ and asserts values: `row_number() OVER (ORDER BY n DESC)` with `ORDER BY id`, on a plain scan and on a derived table.

nodedb/src/bridge/mod.rs
```diff
@@ -15,6 +15,89 @@ pub mod expr_eval {
...
+    ///
+    /// The rows keep the `nodedb_types::Value` codec end to end. Decoding them
+    /// to JSON first downgrades tagged cells — MessagePack binary has no JSON
+    /// representation, so a `BYTES` cell becomes a base64 `String` — and every
+    /// cell the window set does not even touch changes shape with it.
+    ///
+    /// Every step fails loudly: a malformed row, a non-object row, or an
+    /// evaluator column count that does not match the rows is an internal
+    /// error, never a silently skipped row.
+    pub fn evaluate_window_functions_on_msgpack_rows(
```
farhan-syah, line 29: **Blocker 6.** `mod.rs` holds `pub mod` and `pub use` only. Move this adapter to its own file. The right home is `nodedb-query/src/window/` next to `value_eval.rs`: the adapter has no Origin-specific dependency and Lite calls the same evaluator.

nodedb/src/control/server/response_shape/compose.rs
```diff
@@ -120,10 +136,13 @@ pub fn shape_payload_no_plan(
...
         PlanKind::ReturningRows => {
-            ShapeOutcome::Rows(shape_returning_rows(payload, projection, redaction)?)
+            ShapeOutcome::Rows(shape_returning_rows(payload, projection, redaction, None)?)
         }
         PlanKind::SingleDocument | PlanKind::MultiRow => {
-            ShapeOutcome::Rows(shape_generic_rows(payload, projection, redaction)?)
+            // No plan in scope means no statement identity either, so no
+            // sequence stamps: a `nextval(...)` column only ever reaches this
+            // shaper through a streamed/materialized path that carries one.
+            ShapeOutcome::Rows(shape_generic_rows(payload, projection, redaction, None)?)
```
farhan-syah, line 145: **Blocker 3.** Every caller of `shape_payload_no_plan` shapes with `stamper = None`. The `OutputSchema` still carries `sequence: Some(..)` on the `nextval` column. `project_row` finds no cell and emits NULL with success. Reachable now:

- pgwire `routing/set_ops.rs:40`: `SELECT nextval('s') FROM a UNION ALL SELECT id FROM b`
- pgwire `routing/gateway_dispatch.rs:78` and `:231`: every gateway-forwarded read in cluster mode
- pgwire `routing/cluster_array.rs` and `routing/execute_dml_hooks.rs:332`

This is the silent-NULL class #314 closes. The resubmit makes `shape_decoded_rows` return an internal error when a schema column carries `sequence` and no stamper is present. Then it builds the stamper at each call site. All of them have `state`, `database_id`, and `tenant_id`.

nodedb/src/control/server/response_shape/compose.rs
```diff
@@ -247,6 +276,28 @@ pub fn shape_decoded_rows(
             // schemaless collections.
             let columns = derive_columns(&rows);
```
farhan-syah, line 277: Same defect on the star branch. `SELECT *, nextval('s') FROM t` builds a schema with `is_star = true` and a `sequence` column in `schema_from_projection`. `derive_columns` never sees the sequence column. The column is absent from the output, no error. Stamp here too and cover it with a test.

nodedb/src/control/planner/materialized_sum/resolve.rs
```diff
@@ -20,8 +20,102 @@ use super::stored::stored_row_scope;
...
+/// materialized-sum binding with one `PointInsert` task per row.
+///
+/// The page shape is settled per settlement on the Data Plane and per sibling
+/// task on the Control Plane; a page that needs a cross-shard sibling never
+/// completes its Calvin transaction. The per-row shape is exactly what this
+/// server shipped before pages existed, and it settles the same rows, so the
+/// page is un-batched before resolution rather than settled by a new path.
+/// Pages whose bindings are all co-resident (and pages over collections that
+/// drive no binding at all) keep their single task.
+fn un_batch_cross_shard_pages(
```
farhan-syah, line 36: **Blocker 4.** This runs on every `BatchInsert` page in `tasks`, including the page a balanced collection lowers to. `balanced_gate.rs` states the contract: a balanced INSERT is one page so its rows are judged together. Per-row tasks refuse every journal on its first leg. A balanced source that drives a cross-shard sum binding is split into per-row `PointInsert`s here. The contract is gone.

Two more effects of the split:

- the per-row tasks answer one `INSERT 0 1` per row, so #297 stays open for this shape
- the rows no longer commit or fail together

The comment names the root cause: a page with an `ApplyBalanceDelta` sibling never reaches Calvin completion. The resubmit fixes that on the settlement path. It does not re-shape the plan around it.

This file grows from 427 to 531 non-test lines. The hard limit is 500.

nodedb-sql/src/ddl_ast/graph_parse/variants.rs
```diff
@@ -147,30 +174,90 @@ pub(super) fn parse_algo(toks: &[Tok<'_>]) -> Result<NodedbStatement, SqlError>
...
+    cursor: &mut Cursor<'_>,
+    sql: &str,
+) -> Result<NodedbStatement, SqlError> {
+    let collection = cursor
+        .word_after("ON")
+        .or_else(|| cursor.quoted_after("ON"))
         .ok_or_else(|| missing_clause("GRAPH RAG FUSION", "ON <collection>"))?;
-    let params = FusionParams::extract(toks, sql, &RAG_FUSION_KEYWORDS);
+    let params = FusionParams::extract(cursor.tokens(), sql, &RAG_FUSION_KEYWORDS);
+    cursor.consume_rest();
```
farhan-syah, line 208: **Blocker 5.** `consume_rest()` exempts `GRAPH RAG FUSION` from the unclaimed-token check. The comment above says the extractor refuses a malformed option. It does not: every field of `FusionParams::extract` comes from `usize_after`, `quoted_after`, or `word_after`, and an absent keyword is `None`. `GRAPH RAG FUSION ON g QUERY ARRAY[0.1, 0.2] VECTOR_TOPK 5` runs with the default `vector_top_k`. That is the #296 defect on the one statement that opts out.

The resubmit gives the extractor the cursor, so its keywords and values are claimed. It claims the `ARRAY[...]` token span and removes `consume_rest`.

nodedb-sql/src/ddl_ast/graph_parse/cursor.rs
```diff
@@ -0,0 +1,285 @@
...
+        self.word_after(keyword)?.parse().ok()
+    }
+
+    /// The next `count` float words after `keyword`.
+    ///
+    /// Used by the cursor's own tests; the fixed-arity clause families that
+    /// would call it at runtime still read their floats through
+    /// `helpers::array_floats_after`. Kept as the typed accessor the cursor
+    /// tests pin the `claim` bookkeeping to.
+    #[allow(dead_code)] // exercised from `#[cfg(test)]`, so the lib build has no caller
```
farhan-syah, line 127: Dead code in the library build, kept by `#[allow(dead_code)]` for one test. Route `float_pair_after` and `float_triple_after` in `fusion_params.rs` through this accessor, or remove it and its test.

nodedb/src/control/planner/sql_plan_convert/aggregate/plan.rs
```diff
@@ -198,6 +200,71 @@ pub(in crate::control::planner::sql_plan_convert) fn convert_aggregate(
...
+    // catalog path uses above — so the executor receives the body's rows and
+    // evaluates the aggregate arguments / group keys against them. Without
+    // this the aggregate scanned an empty (non-existent) collection and
+    // silently returned NULL / no rows (issue #295).
+    if !matches!(input, SqlPlan::Scan { .. }) {
+        let derived_group_specs = group_by_to_specs(group_by);
+        let derived_agg_specs: Vec<AggregateSpec> =
+            aggregates.iter().map(agg_expr_to_spec).collect();
+        let mut body_tasks = super::super::convert::convert_one(input, tenant_id, ctx)?;
+        if body_tasks.len() == 1
```
farhan-syah, line 216: When the body lowers to more than one task, this falls through to the single-collection path below. The comment above describes that path as a scan of a non-existent collection that returns NULL. Return a typed `PlanError` that names the shape.

nodedb/src/control/planner/sql_plan_convert/aggregate/plan.rs
```diff
@@ -198,6 +200,71 @@ pub(in crate::control::planner::sql_plan_convert) fn convert_aggregate(
...
+                    ),
+                    input: Some(Box::new(body_provider)),
+                    group_by: derived_group_specs,
+                    aggregates: derived_agg_specs,
+                    filters: Vec::new(),
+                    having: having_bytes,
+                    limit,
+                    sub_group_by: Vec::new(),
+                    sub_aggregates: Vec::new(),
+                    grouping_sets: Vec::new(),
```
farhan-syah, line 259: `grouping_sets` is dropped here. `SELECT x, count(*) FROM (SELECT 1 AS x) s GROUP BY ROLLUP(x)` returns the base grouping only, with success. The catalog arm above refuses that case with a `PlanError`. Do the same here.

nodedb/src/control/planner/sql_plan_convert/aggregate/plan.rs
```diff
@@ -198,6 +200,71 @@ pub(in crate::control::planner::sql_plan_convert) fn convert_aggregate(
...
     }
 
+    // Aggregate over a derived/CTE body: the input is not a Scan (a
+    // constant subquery, a set operation, a join materialized earlier), so
+    // there is no per-shard collection to aggregate. Lower the body to a
+    // single coordinator-local ProviderScan sub-plan — the same shape the
+    // catalog path uses above — so the executor receives the body's rows and
+    // evaluates the aggregate arguments / group keys against them. Without
+    // this the aggregate scanned an empty (non-existent) collection and
+    // silently returned NULL / no rows (issue #295).
```
farhan-syah, line 210: Code comments carry no issue numbers. Same at `nodedb-physical/src/physical_plan/query.rs:89`, `sql_plan_convert/expr.rs:375`, `sql_plan_convert/set_ops.rs:326`, and `data/executor/handlers/provider_scan.rs:120`. State the rule in the comment. The issue link lives in the PR.

nodedb/src/control/maintenance/clone_materializer/auto_source.rs
```diff
@@ -0,0 +1,72 @@
...
+                tenant_id,
+                database_id,
+                source_qualified,
+                cursor,
+                system_as_of_ms,
+                txn_id,
+            )
+            .await
+        }
+        EngineType::KeyValue => {
```
farhan-syah, line 54: The kv arm drops `system_as_of_ms` and `txn_id`. `expander_stage.rs` stamps `txn_id` on an in-transaction `INSERT ... SELECT` so the source scan folds the transaction's own writes. A kv source misses rows this transaction inserted, with no error. `AS OF SYSTEM TIME` on a kv source is ignored the same way. Thread both into `kv::scan_source_page`, or refuse with a typed error when either is `Some`.

nodedb/src/control/maintenance/clone_materializer/auto_source.rs
```diff
@@ -0,0 +1,72 @@
...
+                kv::scan_source_page(state, tenant_id, database_id, source_qualified, cursor)
+                    .await?;
+            let entries = pairs
+                .into_iter()
+                .map(|(key, value)| (String::from_utf8_lossy(&key).into_owned(), 0, value))
+                .collect();
+            Ok((entries, next))
+        }
+        // Refused by name instead of silently copied as nothing; these
+        // engines have their own materializers that are not wired here yet.
```
farhan-syah, line 65: `not wired here yet` narrates a future state. State the fact: these engines have no INSERT ... SELECT source materializer. Same for `used to drop it silently` in `strict_format/coerce.rs:241`. Describe what the function does now.

nodedb/src/control/server/native/dispatch/sql_loop.rs
```diff
@@ -204,6 +204,7 @@ pub(super) async fn run_dispatch_loop(
                         database_id,
                         tenant_id: ctx.tenant_id(),
                         redaction: Some(redaction.ctx(&ctx.state.redaction)),
+                        session_sequences: None,
```
farhan-syah, line 207: `ctx.sessions.sequence_values(ctx.peer_addr)` exists for native sessions. Pass it here and at line 319. As written, `currval` after a per-row `nextval` answers on pgwire and raises on native. Same in `native/dispatch/streaming.rs`, where the stamper is built with `None`.

nodedb/src/data/executor/handlers/provider_scan.rs
```diff
@@ -108,7 +112,113 @@ impl CoreLoop {
...
+                                    .to_string(),
+                            },
+                        );
+                    }
+                };
+                for cc in &computed_cols {
+                    if matches!(map.get(&cc.alias), Some(v) if !v.is_null()) {
+                        continue;
+                    }
+                    match cc.expr.eval(&nodedb_types::Value::Object(map.clone())) {
```
farhan-syah, line 162: This block re-implements `apply_projection_msgpack`: decode, evaluate computed columns, encode. It clones the whole row map once per computed column per row. `kv/scan.rs` calls the shared helper for the same job. Call it here.

nodedb/src/control/planner/sql_plan_convert/dml/insert/convert.rs
```diff
@@ -124,45 +132,106 @@ pub(in super::super::super) fn convert_insert(
...
+                            partial: false,
+                            returning: None,
+                            rls_filters: Vec::new(),
+                        }),
+                        post_set_op: PostSetOp::None,
+                        txn_id: None,
+                    });
+                    continue;
+                }
+                if if_absent {
```
farhan-syah, line 154: PostgreSQL answers one `INSERT 0 k` for a multi-row `ON CONFLICT DO NOTHING`. This branch keeps one task and one tag per row, so #297 stays open for that shape. The resubmit gives `BatchInsert` an `if_absent` flag. The Data Plane counts the rows it wrote. The affected count already travels back in the payload.

nodedb-sql/src/planner/select/helpers.rs
```diff
@@ -35,6 +68,13 @@ pub fn convert_projection(
     for item in items {
         match item {
             ast::SelectItem::UnnamedExpr(expr) => {
+                if let Some(sequence) = nextval_sequence_argument(expr) {
+                    result.push(Projection::Sequence {
+                        sequence,
+                        alias: format!("{expr}").to_lowercase(),
```
farhan-syah, line 74: Use `unaliased_projection_alias(expr)` here. The PR introduces it as the single derivation every consumer must use.

nodedb/src/control/planner/sql_plan_convert/expr.rs
```diff
@@ -347,7 +368,18 @@ pub(super) fn inline_cte(plan: &SqlPlan, cte_name: &str, cte_plan: &SqlPlan) ->
                         // offset 0 = unspecified → inherit CTE's offset.
                         offset: if *offset > 0 { *offset } else { *inner_o },
                         distinct: *distinct || *inner_d,
-                        window_functions: inner_w.clone(),
+                        // Window functions: the derived body's own specs run
+                        // first (they produce columns the outer may reference),
+                        // then the outer's. Dropping the outer's here left
+                        // `SUM(n) OVER ...` over a derived table with a Scan
+                        // body silently NULL (issue #295).
```
farhan-syah, line 375: This file grows from 529 to 568 non-test lines. It was already over the 500 hard limit. Move `inline_cte` into its own file.

farhan-syah (changes requested):
**Verdict: closing this PR.** Do not push a fix round here. Resubmit as one PR per issue.

CI is green on `fdba3e378`, so the fmt, clippy, and deny claims hold. The test claim does not: commit `4df41f823` un-registers six wire test modules, and the 16,169-pass figure excludes them.

**Why one batch cannot be fixed in place**

| Signal | Consequence |
|---|---|
| Six test modules removed in commit 5 of 29, undisclosed | Every validation claim in the PR needs a re-audit of all 29 commits. |
| Window ORDER BY fix lives in an evaluator no Origin path calls | A later commit moved both callers and nobody re-checked the earlier fix. |
| `un_batch_cross_shard_pages`, `consume_rest()`, `grouping_sets: Vec::new()` | Three workarounds around root causes. |
| #317 ships with no test and its existing test module removed | Redo, not patch. |

**Blockers found** (inline comments carry each one)

| # | Where | Defect |
|---|---|---|
| 1 | `nodedb/tests/wire/cases/mod.rs` | Six wire test modules un-registered: `sql_convert_column_defs`, `sql_declared_column_types`, `sql_default_vector_primary`, `sql_default_volatility`, `sql_typeguard_default_gate`, `timeseries_join_time_rendering`. The files stay on disk. They cover #317, #318, and `nextval` DEFAULT volatility. |
| 2 | `nodedb-query/src/window/eval.rs`, `nodedb/src/bridge/mod.rs` | The ORDER BY fix is in `evaluate_window_functions`. Both Origin paths call `evaluate_window_functions_value`, which evaluates partitions in arrival order. The two new wire tests use one key for both orders, so they pass either way. |
| 3 | `response_shape/compose.rs`, pgwire `set_ops.rs`, `gateway_dispatch.rs` | No stamper on any `shape_payload_no_plan` door or on the `SELECT *` branch. A `nextval(...)` column answers NULL with success on `UNION`, cluster gateway reads, and `SELECT *, nextval('s') FROM t`. |
| 4 | `materialized_sum/resolve.rs` | `un_batch_cross_shard_pages` splits every `BatchInsert` page, including a balanced collection's page. The split rows answer `INSERT 0 1` per row and lose atomicity. |
| 5 | `graph_parse/variants.rs` | `GRAPH RAG FUSION` calls `consume_rest()`. `FusionParams::extract` validates nothing. A mistyped option defaults. |
| 6 | `nodedb/src/bridge/mod.rs` | A 70-line function in a `mod.rs`. `mod.rs` holds `pub mod` and `pub use` only. |

**Should-fix found**

| # | Where | Item |
|---|---|---|
| 7 | `aggregate/plan.rs` | Derived-body aggregate falls through silently past one task. `grouping_sets` dropped silently. |
| 8 | `clone_materializer/auto_source.rs` | The kv arm drops `txn_id` and `system_as_of_ms`. |
| 9 | five files | `issue #295` in code comments. `used to` and `not wired here yet` narration. |
| 10 | native `sql_loop.rs` | `session_sequences: None` while the native session map exists. `currval` diverges between pgwire and native. |
| 11 | `provider_scan.rs` | Re-implements `apply_projection_msgpack`. Clones the row map per computed column per row. |
| 12 | `dml/insert/convert.rs` | `ON CONFLICT DO NOTHING` multi-row INSERT answers one tag per row. |
| 13 | `graph_parse/cursor.rs` | `floats_after` is dead code behind `#[allow(dead_code)]`. |
| 14 | `materialized_sum/resolve.rs`, `sql_plan_convert/expr.rs` | `resolve.rs` grows 427 to 531 non-test lines. Hard limit is 500. `expr.rs` grows 529 to 568. |
| 15 | `select/helpers.rs` | `Projection::Sequence` alias bypasses `unaliased_projection_alias`. |
| 16 | tests | No wire test for #297, #311, #317, or the #318 refusals. |

**What is correct and can return as standalone PRs, each with its test**

- #296 graph cursor, without the RAG FUSION exemption
- #306 strict `VECTOR` element coercion and `BadRequest` preservation
- #318 typeguard `database_id` threading, CONVERT primary-key carry, cross-column refusal
- #305 derived `distance` column
- #297 kv `INSERT 0 1` tag and the `numeric_code_to_sqlstate` mappings
- constant-row `cell_keys` fix

**What needs a redesign, not a patch**

- #314: `shape_decoded_rows` must fail when a schema column carries `sequence` and no stamper is present. Every door must carry one.
- #295: ordering in `evaluate_window_functions_value`, one evaluator. Test with outer order different from window order.
- #297 multi-row: `BatchInsert` with `if_absent` and a Data Plane count. Fix the page and `ApplyBalanceDelta` Calvin completion. No un-batching.
- #317: re-register `timeseries_join_time_rendering`, make it pass, add join and star coverage.

**Conditions for every resubmit**

1. One issue per PR.
2. The PR carries the test that fails on `main` without the fix.
3. No test module removed or ignored without a sentence in the PR body that says why.

**Ownership**

| Issue | Owner |
|---|---|
| #295, #297, #314, #317 | maintainers (the redesign items above) |
| #296, #305, #306, #311, #318 | open for your resubmit, one PR each |

The assignee field on each issue is the source of truth.

Satellite note: `Projection::Sequence` is a new `nodedb-sql` enum variant. `nodedb-lite`'s `execute_plan` must handle it before the next crate publish.

I did not run the six un-registered modules against this branch.

PR #326, opened 2026-09-16
fix(pgwire): answer kv writes with one INSERT tag

EnRaiha:
## Why

Key-value writes answered a bare `OK` command tag: an `INSERT` or `PUT` against a kv collection reported no affected count, so drivers saw an empty tag where `INSERT 0 1` is expected. The count-bearing tag semantics are the kv half of #297.

## What

- `data/executor/handlers/kv/crud/write_basic.rs`: kv `Put` and `Insert` answer through `response_affected(task, 1)` — the write reports the row it wrote instead of a bare OK.
- `response_shape/types/plan_kind.rs`: `KvOp::Insert`, `KvOp::Put`, and `KvOp::BatchPut` classify as `DmlResult("INSERT")`, so the routed response renders `INSERT 0 1` rather than an execution tag.
- The wire case extends `sql_dml_affected_counts` with the kv INSERT/PUT counts.

Refs #297.

## Validation

- `cargo check -p nodedb --all-targets` — clean (rebased on current `main`).
- `cargo test -p nodedb --test wire sql_dml_affected_counts` — 16 passed.
- Repository preflight passes.

## Notes

- The earlier revision of this PR also carried the routed shaper SQLSTATE work; that content now lives in its own PR (#341).


farhan-syah:
Superseded by #356, which carries the kv INSERT/PUT tag and count fix along with the rest of #297 (one folded tag per statement, real counts on every engine and protocol, transaction overlays, and the cluster fold). Closing this one; thanks for the groundwork.

PR #327, opened 2026-09-16
fix(sql): keep one cell per duplicate constant column

EnRaiha:
### Summary

`SELECT nextval('a'), nextval('a')` legally repeats an output name. The constant-result row is a JSON object keyed by column name, so the second cell overwrote the first and both wire columns rendered the last value. The payload and the output schema's lookup keys now use the same unique per-column keys every response encoder derives.

### Red proof (the test fails on `main` without the fix)

```
────────────
     Summary [  10.891s] 1 test run: 0 passed, 1 failed, 10256 skipped
  TRY 2 FAIL [   4.311s] (1/1) nodedb::wire cases::sql_sequences::duplicate_constant_column_names_keep_their_own_cells
error: test run failed
```

### Validation (local, on this branch)

| Check | Result |
|---|---|
| the new test, with the fix | 1 passed |
| related suites: `sql_sequences` + `cell_keys` + `output_schema` | 53 passed |
| `cargo fmt --all` | no change |
| `cargo clippy -p nodedb --all-targets --all-features --profile ci -- -D warnings` | clean |
| repository hygiene gate (test registry, comment and commit hygiene, module contracts) | pass |

One commit; touches `sql_plan_convert/output_schema/build.rs`, `sql_plan_convert/set_ops.rs`, and the new wire test only.


farhan-syah (approved):
Approve.

Verified statically against the branch ref (no checkout):

| Claim | Result |
|---|---|
| `cargo fmt --all` no change | Confirmed. `rustfmt --check` clean on all three files. |
| One commit, three files | Confirmed. `22a0f42`, +59 / −13. |
| Test fails on `main` | Consistent with the code. On `main` both cells land under `nextval`, so the row holds one value and both columns read it. |
| Comment hygiene | Clean. No issue numbers, links, or pre-fix narration. |

Layer is correct. Both producers of the constant-result shape (`set_ops.rs` payload, `build.rs` schema) now derive keys from `cell_keys`, the one function every encoder already reads through. No fallback, no second accepted shape.

Out of scope for this PR, tracked separately: the extended-query path (`pgwire/handler/prepared/execute.rs`) rebuilds its projection from the Describe fields with `lookup_key = field name`, and that projection overrides the planner schema (`routing/execute.rs`). `SELECT nextval('s'), nextval('s')` over Bind/Execute still collapses, as does `SELECT w.id, b.id`. That defect predates this PR.

PR #329, opened 2026-09-16
Accept ARRAY vector literals and typed value errors

### LINKED ISSUE #306: ARRAY vector literal loses every element on document_strict INSERT, reported as 0 elements under XX000

EnRaiha:
# Summary

`ARRAY[0.1, 0.2, 0.3]` folds its literals to `Decimal` in the planner. Strict element coercion kept `Float` and `Integer`, dropped every other element, and reported the literal as having 0 elements — and that mismatch surfaced as `XX000`, an internal error, for what is a client value problem. A non-numeric element was equally opaque: no element index, no type, and the same internal class.

This PR fixes the coercion and the class. Elements of every numeric shape are read; a bad element is refused by index and type; and a client value error keeps its `BadRequest` class across the Data Plane boundary instead of degrading to `XX000`.

## Behaviour changes

| Path | Before | After |
|---|---|---|
| `INSERT … ARRAY[0.1, 0.2, 0.3]` into `VECTOR(3)`, strict schema | `XX000`, "got 0 elements" | the row inserts |
| `ARRAY[0.1, 'nope', 0.3]` | `XX000` | `42601`, naming the element index and its type |
| `ARRAY[NaN]` / `ARRAY[Inf]` | folded through | refused |
| A client value error crossing the Data Plane or the cluster wire | `XX000` | `42601` |

Closes #306.

# Root cause

- `extract_vector_floats` in `strict_format/coerce.rs` matched `Value::Float` and `Value::Integer` only. The planner's literal folding produces `Decimal`, so a literal `ARRAY[0.1, 0.2, 0.3]` contributed zero elements and the column reported `got 0 elements`.
- The follow-up problem was the error class, not the coercion: `crate::Error::BadRequest` had no mirror in the bridge's `ErrorCode` enum, so the conversion fell through the catch-all to `ErrorCode::Internal { detail }` — every client value error crossed the Data Plane boundary as `XX000`, even though the direct (single-node) path already assigned `42601`.
- Two strict-encoder wrap sites (`handlers/convert.rs`, `point/update/post_image.rs`), the update-from-join collect pass, and two doors found by the audit rounds (below) had the same fold-into-`Internal` shape.

# What changed

Commit `0fd44a626` — coercion:

- `extract_vector_floats` reads `Float`, `Integer`, `Decimal`, and numeric `String` elements; a non-numeric element is rejected with its index and type; non-finite values are refused with the value.
- The four `binary_tuple` wrap sites (`stored_body.rs`, `stage_write/body.rs` ×2, `stage_upsert.rs`) preserve `BadRequest` instead of wrapping it.

Commit `ddefcaa0f` — wire coverage:

- `document_strict_inserts_a_decimal_array_literal_into_a_vector_column` (the filed case)
- `document_strict_reports_a_non_numeric_element_by_index_and_type` (message names the element; SQLSTATE pinned)

Commit `6c29ac9fd` — the bridge:

- `ErrorCode::BadRequest { detail }` added to the bridge envelope enum with the same doc-comment pattern as its siblings, and preserved in `From<crate::Error>`.
- Mirrored on the cluster wire (`DataPlaneErrorCode`, appended variant; the enum's own doc says variants are appended, never reordered) with both round-trip conversions.
- pgwire renders it `42601` (the code the direct path already assigns). The node-abort classifier treats it as *not established*: the same code can be raised after an earlier step of a multi-step plan applied, so refusal-without-apply is not provable from the code alone.

Fixes from the audit rounds:

- `f912fb2b1`: two more doors preserved `BadRequest` — `handlers/convert.rs` (CONVERT row encode) and `point/update/post_image.rs` (strict re-encode). Tests added: the abort classifier covers `BadRequest`; the cluster wire round-trips it verbatim.
- `e3e129563`: the update-from-join collect pass (`handlers/update_from_join.rs`), which encodes strict post-images, preserved it too.

## Regression proof

On base `main`, without this change: the filed case reports `expected VECTOR(3), got 0 elements (SQLSTATE XX000)`, and the bad-element case finds no element message. With this change both pass, and the bad-element SQLSTATE is asserted as `42601`.
# Tested

- `cargo nextest run -p nodedb --test wire -E 'test(cases::engine_surface_document_strict)'` — 2 new cases pass
- error-path regressions: **42 passed** across `error_from_data_plane`, `data_plane_error_wire`, `write_abort`, and the sqlstate/error-code unit tests
- convert/typeguard modules: **22 passed** (`sql_convert_column_defs`, `sql_typeguard_default_gate`, `sql_typeguard_defaults`)
- `cargo clippy -p nodedb -p nodedb-cluster --lib -- -D warnings` — clean
- preflight — pass

## Review

A separate, read-only parity audit ran over this branch before submission. It found the same defect class on further sites and asked for two tests; all are fixed and verified here.

| Finding | Resolution |
|---|---|
| Two strict-encode doors folded `BadRequest` into `Internal` (`handlers/convert.rs`, `point/update/post_image.rs`) | both preserve the class; the abort classifier and the cluster wire round-trip are covered by tests (`f912fb2b1`) |
| The update-from-join collect pass folded it too | preserved on that path; targeted tests pass (`e3e129563`) |
| Cross-version compatibility of the appended wire variant | the enum's append-only discipline; CI covers the workspace build |
# Notes

- Host note: on this machine the native wire tests overflow the default thread stack and need `RUST_MIN_STACK=16M`. The same abort reproduces on an unrelated branch with no native-path changes, so it is a host property, not this change. CI runs them at the repository default.
- The wire variant is appended and never reordered; existing slots are untouched.

# Commits

- `0fd44a626` — `fix(strict-format): coerce ARRAY vector literals into a strict VECTOR`
- `ddefcaa0f` — ARRAY vector literal wire coverage
- `6c29ac9fd` — `fix(bridge): carry a bad request across the Data Plane boundary`
- `f912fb2b1` — `fix(executor): keep a bad request out of XX000 on the convert and post-image doors`
- `e3e129563` — `fix(executor): keep a bad request typed on the update-from-join collect pass`


farhan-syah:
Closing. `extract_vector_floats` copies `vector_string::floats_from_value`, which already reads Float/Integer/Decimal/String elements with typed errors. Strict coercion must call that helper. A bad element is a data error, not a syntax error (`42601`). The maintainers will fix #306 in-tree.

PR #331, opened 2026-09-16
Resolve typeguards per database and carry CONVERT keys

### LINKED ISSUE #318: A cross-field typeguard guard passes the DEFAULT gate at declaration and fails at every insert

EnRaiha:
# Summary

Three defects on the typeguard and CONVERT declaration paths, and one gate that all of them pass through:

1. **`CREATE TYPEGUARD` resolved against the wrong database.** The handlers read `DatabaseId::DEFAULT` while collections live under the session database, so a guard declared in another database answered `42P01` for a collection that exists.
2. **`CONVERT` dropped the source primary key.** The conversion rebuilt the strict schema from the column list, and the key column lost its primary-key mark; every later insert failed with "no resolved primary key". A column list that omits the key column is now refused.
3. **A cross-field guard carried onto a column DEFAULT failed at every insert.** A guard `VALUE LOWER(status)` is evaluated per row; carried onto a strict column it becomes a `DEFAULT`, which is const-folded once with no row in scope — the CONVERT was accepted and the first insert failed with `UnevaluableDefault`. It is now refused at the declaration, naming field and clause.

The same invariant now holds at one gate (`validate_constant_clause_expr`) for every producer of a column `DEFAULT`: `CREATE COLLECTION`, a `CONVERT` column list, a `CONVERT` typeguard carry, and `ALTER … ADD COLUMN`.

**Scope statement (requested by the audit).** The issue text names defect 3 only. This PR also carries defects 1 and 2, which the maintainer's resubmission conditions grouped under #318 — typeguard `database_id` threading, CONVERT primary-key carry, and the cross-column refusal. They ride together as in-PR prerequisites of the same declaration path; they are stated here rather than split, per the audit's remedy.

## Behaviour changes

| Path | Before | After |
|---|---|---|
| `CREATE TYPEGUARD` in a session database | `42P01` | declared in that database |
| `INSERT` after `CONVERT … (id TEXT, v TEXT)` | "no resolved primary key" | inserts; the source key is carried |
| `CONVERT` column list that omits the source key | accepted, the key lost | `42601` |
| `CREATE` / `CONVERT` / `ALTER ADD COLUMN … DEFAULT LOWER(other)` | accepted; the first insert failed `UnevaluableDefault` | `42601` at declaration |
| A guard `VALUE LOWER(status)` carried by `CONVERT` | accepted; the first insert failed | `42601` at the `CONVERT`, naming field and clause |
| `ALTER … ADD COLUMN is_default BOOLEAN` | accepted, the name mis-read as a `DEFAULT 'BOOLEAN'` clause | accepted, no clause read |

Closes #318.

# Root cause

- **Database resolution:** `typeguard/handlers.rs` built every request with `DatabaseId::DEFAULT`; `validate.rs` did the same when resolving the target collection. The session's database never reached the handler.
- **Primary key carry:** `convert/driver.rs` built the target schema from the explicit column list alone. The source's key column (`primary_key`) was not consulted, so the converted schema had no key and `assign_target_surrogate` failed on the next insert with "no resolved primary key".
- **Cross-field guard:** `CompiledDefault::declare` classifies a `DEFAULT`; a function call like `LOWER(status)` parses to an `Expr` and was accepted. Nothing checked whether the expression names a column. The trap appeared only at the first insert (`UnevaluableDefault`).
- **The class:** the reference check lived at two CONVERT call sites, so `CREATE COLLECTION` (`collection/create/build.rs`) and a `CONVERT` column list (`convert/column_defs.rs`) still accepted a cross-column `DEFAULT` and failed at the first insert. The audit found `ALTER … ADD COLUMN` as a fourth producer that bypassed the gate entirely.

# What changed

Commit `1b7dbd907` — the three defects:

- `database_id` threaded through the seven typeguard handlers and `validate_typeguard`; the router resolves the collection in the session database.
- `CONVERT` marks the source key column in the converted schema; a column list that omits it is refused with `42601`.
- `nodedb-sql` exposes `default_expr_references_columns` — the classifier's own parse, no second name list — and `CONVERT` refuses a carried guard whose expression names another column.

Commit `780176fac` — the explicit-list path: a guard the list covers and that names another column is refused there too (the list defines the schema and carries no guard onto it).

Commit `879b1329f` — one gate:

- `validate_constant_clause_expr(clause, owner, expr)` is the single place that refuses an unregistered function name, an unparsable expression, or a column reference; `validate_column_default` becomes the `DEFAULT`-clause form of it, and the two CONVERT-local copies of the reference check are deleted.
- `default_expr_references_columns` now mirrors `classify`: a generator (`UUID_V7`, `gen_uuid_v7()`) or a literal/parametric form carries no expression and is never read as a column reference. (The first cut parsed the raw text and refused `DEFAULT UUID_V7` — caught by the existing `sql_default_expressions` and `sql_typeguard_defaults` modules.)

Commit `11a3c7b26` — `ALTER … ADD COLUMN`: the declared definition passes the same gate.

Commit `82ea86a91` — wording: the carry path passes the clause through, so a guard `VALUE` reports `VALUE`, not `DEFAULT` (audit advisory).

Commit `b9bc243fb` — the gate reads the parsed `DEFAULT`, never the definition text, and the type parser's clause search requires the keyword to start a token: a column named `is_default` is no longer read as a `DEFAULT` clause.

Merge `44b21159a` — folds `origin/main` (#336 instant typing, #327 constant-row cell keys) into
  the branch, resolving the DDL-gate conflict by composing both sides: `validate_column_default_clause`
  runs the constant/column-reference check and the literal type/range coercion at one gate.

## Regression proof

On base `main`, without this change: the insert after `CONVERT` fails with "no resolved primary key", a column list that omits the key is accepted, the cross-column guard is accepted, `CREATE TYPEGUARD` in a session database answers `42P01`, and a cross-column `DEFAULT` is accepted at `CREATE`/`CONVERT`/`ALTER` ("got success"). With this change each is refused or resolved as listed in Behaviour changes.
# Tested

- `cargo nextest run -p nodedb --test wire -E 'test(cases::sql_convert_column_defs) | test(cases::sql_default_expressions) | test(cases::sql_typeguard_default_gate) | test(cases::sql_typeguard_defaults)'` — **41 passed**
- unit: `only_the_expression_branch_can_reference_a_column` pins the classifier mirroring (generators and literals are not references; a column and a concatenation are)
- audit's own runs: 23/23 touched-module tests, 228/228 blast-radius tests
- `cargo fmt --all -- --check`, `cargo clippy -p nodedb -p nodedb-sql --lib -- -D warnings`, preflight: clean

## Review

A separate, read-only parity audit ran over this branch before submission. Its findings and their resolutions:

| Finding | Resolution |
|---|---|
| The reference invariant was enforced at two `CONVERT` call sites, not the gate every producer calls | one gate (`879b1329f`); `CREATE`, `CONVERT`, the explicit-list guard path and `ALTER` all route through it |
| `ALTER … ADD COLUMN … DEFAULT` bypassed the gate | gated (`11a3c7b26`) |
| The `ALTER` gate read the whole definition, so a name containing `default` was read as the clause | the gate reads the parsed default, and the parser's clause search requires the keyword to start a token (`b9bc243fb`) |
| Clause wording: a carried `VALUE` reported `DEFAULT` | the clause is passed through (`82ea86a91`) |
| The issue text names one of the three defects this PR carries | scope statement above; recorded on the issue |
# Exclusions

- The guard-carry path still ignores a guard whose field the explicit list does not mention; that guard is not part of the target schema at all (unchanged behaviour).
- The maintainer-owned paths (#295, #297, #314, #317) are not touched.

# Commits

- `1b7dbd907` — `fix(typeguard): resolve in the session database and keep the CONVERT source key`
- `780176fac` — `fix(convert): refuse a cross-column guard on the explicit-list path`
- `879b1329f` — `fix(ddl): refuse a column DEFAULT that names another column at the one gate`
- `11a3c7b26` — `fix(ddl): gate an ALTER ADD COLUMN DEFAULT like a CREATE column DEFAULT`
- `82ea86a91` — `fix(ddl): name the clause the author wrote on the carried guard`
- `b9bc243fb` — `fix(ddl): read an ALTER default from the parsed definition`
- `88f2edaa1` — `style: rustfmt`


farhan-syah:
Closing. `expr_references_column` is a new recursive column walker, while its own doc says no second walker is written. `nodedb_sql::types::plan::referenced_columns` (`expr_scan.rs`) already does this. The PR also bundles three defects. The maintainers will fix #318 in-tree on the existing walker.

PR #332, opened 2026-09-16
Declare search cells over a closed derived schema

### LINKED ISSUE #305: SEARCH synthetic columns are unresolvable from a derived table over a closed-schema collection

EnRaiha:
# Summary

A derived relation over a closed-schema source refused `s.distance` with `42703`, while the same projection over an open-schema source ran. The response layer answers a search-shaped query with two synthetic cells the schema cannot name: `distance` (Float64) and `_surrogate` (the internal row id, resolved to the user primary key). The declaration follows the plan: `SqlPlan::carries_search_cells()` matches the three variants whose rows lower to the vector-hit shape (`VectorSearch | SparseSearch | MultiVectorSearch`), and `infer_subquery_relation` reads the cells from the subquery's own plan — the source the planner already uses, never a second AST reading of it.

Closes #305.

## Behaviour changes

| Path | Before (main) | After |
|---|---|---|
| `s.distance` over a closed-schema derived relation whose body routes to a search (`ORDER BY vector_*` and the operator forms `<=>` / `<#>` / `<->`, `sparse_score`, routing `WHERE` forms, `WHERE multi_vector_search(...)`) | `42703` | resolves; every returned hit carries a numeric value |
| `s._surrogate` over the same shapes | `42703` | resolves to the integer id |
| Shapes the planner sends to scalar evaluation (a comparison wrapped around a call, a one-argument `ORDER BY vector_distance(..)`, a body whose `FROM` is neither `Scan` nor `Join`) | `42703` | `42703` — a declared cell is never NULL |
| Queries that previously ran | unchanged | unchanged |

# Root cause

Derived-relation inference (`resolver/derived.rs`) re-derived "is this a search?" from the AST. The planner answers the same question in `where_search` / `order_by/triggers`, and the two diverged in both directions: declarations for shapes that route no search (silent NULL cells), and no declaration for `SqlPlan::MultiVectorSearch`.

# What changed

Rebased onto `a7f597257`; two commits.

Commit `e9f969f60` — the fix: `SqlPlan::carries_search_cells()` matches the three variants whose rows lower to the vector-hit shape; `infer_subquery_relation` takes the subquery's plan (`derived_from` passes `inner_plan`, `entry` passes `cte_plan`, `TableScope`'s derived arm plans non-lateral factors for the same input — a correlated `LATERAL` factor cannot be planned at scope time and routes no cells). The four AST heuristic functions and the third registry singleton are deleted.

Commit `e0a51d035` — the tests: the filed case (closed schema plus a vector index, `SEARCH … USING VECTOR(…)` in the derived body), the pinned outcomes for the previously tolerant assertions, and guards for the four divergences (`sparse_score` in `WHERE`, one-argument `ORDER BY vector_distance`, a body off a derived relation, and `WHERE multi_vector_search`).

## Regression proof

On `main` without the change, the reworked wire case fails 5 of 20 — including the filed `SEARCH … USING VECTOR(…)` case and the `_surrogate` projection. On the branch head all 20 pass.

# Tested

- `cargo nextest run -p nodedb-sql --lib` — 892 pass (derived-relation and `carries_search_cells` unit tests included)
- `cargo nextest run -p nodedb --test wire -E 'test(cases::sql_search_subquery_composition)'` — 20 pass; on `main` with the same tests: 15 pass / 5 fail
- lateral, recursive CTE, insert-select, vector-index and hybrid modules — 52 pass
- `cargo fmt --all -- --check`, `cargo clippy -p nodedb-sql --all-targets`, `nodedb-preflight.sh`: clean

# Notes

- `synthetic_column` declares the cells as nullable `Unknown` type: the declaration resolves the name; the response layer supplies the value.
- `WHERE multi_vector_search(...)` plans `SqlPlan::MultiVectorSearch`, which the lowering step does not support yet (`42601`, variant named). The cell declaration is plan-derived, so that refusal is the honest one; `42703` no longer appears for it.
- A body off a derived relation gives `order_by/triggers` no `Scan`/`Join` to read, so it routes no search; its projection refuses the cells.


nodedb-sql/src/resolver/derived.rs
```diff
@@ -86,7 +86,103 @@ fn infer_projection(
...
+
+    // The SEARCH preprocessor rewrites `SEARCH c USING VECTOR(...)` into
+    // A search-shaped body answers with the source columns plus two synthetic
+    // cells: `distance` (Float64) and `_surrogate`, the internal row id
+    // resolved to the user primary key. Neither is a declared column of a
+    // closed-schema source, so derived-relation inference must declare them:
+    // without it `s.distance` and `s._surrogate` resolve against no relation
+    // and are refused with 42703, while the same projection over an
+    // open-schema source runs.
+    if order_by_triggers_a_search(query) || where_triggers_a_search(&query.body) {
```
farhan-syah: **Blocker: wrong layer.** This re-derives the planner's routing from the AST. `planner/select/derived_from.rs` computes `inner_plan = plan_query(subquery, …)` and on the next line calls `infer_subquery_relation` with the AST instead of the plan. `planner/select/entry.rs` does the same for a CTE (`cte_plan` then `infer_subquery_relation`).

The plan already answers the question. The variants whose rows carry `distance` + `_surrogate` are the ones that lower to a vector-hit shape (`exchange/resolve/exchange/post_process_arm.rs`, `classify_hit_shape`): `SqlPlan::VectorSearch | SparseSearch | MultiVectorSearch`.

The AST copy diverges today:

| Query | Planner | This resolver | Result |
|---|---|---|---|
| `WHERE sparse_score(…) > 0.1` | scalar fallback (`where_search.rs`, `dispatch_trigger`) | declares | `s.distance` is NULL where `main` raised `42703` |
| `ORDER BY vector_distance(x)` (one arg) | `Ok(None)` (`order_by/triggers.rs`, `args.len() < 2`), scan | declares | NULL |
| `WHERE multi_vector_search(…)` | `SqlPlan::MultiVectorSearch` (`where_search.rs`), vector-hit rows | not declared | `42703` stays |
| body whose FROM is not `Scan` / `Join` | `Ok(None)` (`try_extract_sort_search`) | declares | NULL |

The first two turn a loud error into a silent NULL: the same class #305 was filed against, one layer up.

End state: one predicate on `SqlPlan` (`fn carries_search_cells(&self) -> bool` matching those three variants). `infer_subquery_relation` takes the inner plan. `expr_triggers_a_search`, `expr_contains_a_search`, `order_by_triggers_a_search`, `where_triggers_a_search` go away. `resolver/columns.rs` (`TableScope`, the `Derived` arm) also infers without a plan and needs the same input.

nodedb-sql/src/resolver/derived.rs
```diff
@@ -86,7 +86,103 @@ fn infer_projection(
     catalog: &dyn SqlCatalog,
     query: &ast::Query,
 ) -> Result<(Vec<ColumnInfo>, bool)> {
-    infer_body(catalog, &query.body)
+    let (mut columns, open) = infer_body(catalog, &query.body)?;
+
+    // The SEARCH preprocessor rewrites `SEARCH c USING VECTOR(...)` into
```
farhan-syah: **Blocker: broken comment.** This sentence ends here. The next line starts a different one. Remove this line or finish it.

nodedb-sql/src/resolver/derived.rs
```diff
@@ -86,7 +86,103 @@ fn infer_projection(
...
+fn order_by_triggers_a_search(query: &ast::Query) -> bool {
+    query.order_by.iter().any(|order| match &order.kind {
+        ast::OrderByKind::Expressions(exprs) => exprs
+            .iter()
+            .any(|ordered| expr_triggers_a_search(&ordered.expr)),
+        ast::OrderByKind::All(_) => false,
+    })
+}
+
+/// WHERE routing covers the vector trigger only: `where_search`'s dispatch
```
farhan-syah: **Blocker: doc contradicts code.** "covers the vector trigger only", but `expr_contains_a_search` → `expr_triggers_a_search` matches `SparseSearch` too. The planner's WHERE dispatch sends `SparseSearch` to scalar evaluation, so the code is the wrong side of this contradiction, not the comment. Resolved by the layer change above.

nodedb-sql/src/functions/registry.rs
```diff
@@ -121,6 +121,13 @@ pub struct FunctionRegistry {
     functions: Vec<FunctionMeta>,
 }
 
+/// The process-wide built-in registry. Read-only after construction; used
+/// where a function name alone decides routing.
+pub fn builtin_registry() -> &'static FunctionRegistry {
```
farhan-syah: **Should-fix: third registry singleton.** `planner/const_fold.rs` (`DEFAULT_REGISTRY`) and `resolver/expr/functions.rs` (`FUNCTION_REGISTRY`) already hold a `LazyLock<FunctionRegistry>`. The layer change removes this caller. If a static stays, keep one and point the other two at it.

nodedb/tests/wire/cases/sql_search_subquery_composition.rs
```diff
@@ -264,3 +264,230 @@ async fn outer_order_by_distance_then_limit_takes_farthest() {
...
+        .query_text(
+            "SELECT s.distance FROM \
+             (SELECT * FROM sp_cells \
+              WHERE vector_cosine_distance(embedding, ARRAY[0.1, 0.2, 0.3]) < 2.0) s",
+        )
+        .await
+        .unwrap_or_else(|e| panic!("a WHERE-clause trigger must declare s.distance: {e}"));
+    for row in &rows {
+        assert!(
+            row.is_empty() || row.parse::<f64>().is_ok(),
```
farhan-syah: **Should-fix: the test certifies a NULL cell.** Once the declaration comes from the plan, a declared cell is always filled. Pin `parse::<f64>()` with no `is_empty()` escape. Same at the sparse case below.

nodedb/tests/wire/cases/sql_search_subquery_composition.rs
```diff
@@ -264,3 +264,230 @@ async fn outer_order_by_distance_then_limit_takes_farthest() {
...
+    let sql = "SELECT s.distance FROM \
+               (SELECT * FROM sp_sparse ORDER BY sparse_score(terms, '{3: 1.0}') LIMIT 2) s";
+    match server.query_text(sql).await {
+        Ok(rows) => assert!(
+            rows.iter()
+                .all(|r| r.is_empty() || r.parse::<f64>().is_ok()),
+            "distance must be numeric where filled: {rows:?}"
+        ),
+        Err(e) => assert!(
+            !e.contains("42703"),
```
farhan-syah: **Should-fix: the `Err` arm asserts nothing about the fix.** Any error without `42703` passes. Same shape at the no-index case above (`!e.to_string().is_empty()`). With the plan-derived declaration this query either resolves with numeric cells or the sparse read refuses for a stated reason. Pin one outcome per test.

farhan-syah (changes requested):
Request changes.

Checked statically against the branch ref (no checkout):

| Claim | Result |
|---|---|
| `cargo fmt --all -- --check` clean | Confirmed on every changed file. |
| Comment hygiene | One broken comment (inline). No issue numbers or links. |
| "every trigger and every shape" | Not true. The resolver re-derives routing from the AST and diverges from the planner in both directions (inline, blocker). |
| Commit list | The body cites `cdb2e71a1`, `4eddcfeb5`, `235b7a28b`, `994372149`. The branch holds `bec78889`, `15cea2e`, `ca7baec`, `eeb56bb`. Update the body after the rebase. |

Blockers are the layer finding, the broken comment, and the doc/code contradiction. The registry singleton and the tolerant tests are should-fix; both fall out of the layer change.

What landed correctly: the closed-schema wire case reproduces #305, and `_surrogate` is now declared beside `distance`.

EnRaiha:
Addressed and pushed — branch head `2a51f66ce` (layer move `995215c8c`, then merged with `origin/main`).

- **Wrong layer (blocker):** the decision now comes from the plan. `SqlPlan::carries_search_cells()` matches `VectorSearch | SparseSearch | MultiVectorSearch`; `infer_subquery_relation` takes the subquery's plan — `derived_from.rs` passes `inner_plan`, `entry.rs` passes `cte_plan`, and `TableScope`'s derived arm plans non-lateral factors so it gets the same input. The four AST heuristics are deleted. A correlated LATERAL factor cannot be planned at scope time, so it carries no plan and routes no cells (stated in the body).
- **Broken comment:** it sat inside the block the layer change deletes.
- **Doc contradiction:** resolved by the source of truth, not the comment. WHERE sparse routing is the planner's `Ok(None)` and now declares nothing.
- **Third registry singleton:** `builtin_registry()` deleted; its only caller was the heuristic.
- **Tolerant tests:** `:459` pins a numeric cell with no empty escape; `:489` and the no-index case pin one outcome each. The divergence rows are pinned as guards: sparse `WHERE` → `42703`, one-argument `ORDER BY vector_distance(..)` → refusal, body off a derived relation → refusal, `WHERE multi_vector_search(..)` → declared (lowering refuses `42601`, never `42703`).
- **Commit list:** body updated (`bec788896`, `15cea2eec`, `ca7baec72`, `eeb56bb4e`, `995215c8c`; merge `2a51f66ce`).

Red proof: on `main` with the reworked case, 5 of 20 fail — including the filed `SEARCH … USING VECTOR(…)` case and the `_surrogate` projection; branch head 20/20 pass. `nodedb-sql` unit tests 921 pass; lateral, recursive-CTE and insert-select modules 52/52; preflight and fmt clean.

farhan-syah:
Rebase, and force with release. Rebuild the commits

EnRaiha:
Rebased onto `a7f597257` and the history is rebuilt: `e9f969f60` (fix) then `e0a51d035` (tests). Force-pushed with lease.

The rebuilt tree is identical to the previously audited `2a51f66ce` (`git diff 2a51f66ce e0a51d035` is empty), so the parity audit still covers the content. Preflight passes on the rebuilt head.

nodedb-sql/src/resolver/derived.rs
EnRaiha: Fixed in `e9f969f60`: the decision comes from the plan — `SqlPlan::carries_search_cells()` matches `VectorSearch | SparseSearch | MultiVectorSearch`; `infer_subquery_relation` takes the inner plan (`derived_from` passes `inner_plan`, `entry` passes `cte_plan`), and `TableScope`'s derived arm plans non-lateral factors for the same input. The four AST heuristics are deleted.

nodedb-sql/src/resolver/derived.rs
EnRaiha: The stale sentence sat inside the block the layer change removes; it is gone in `e9f969f60`.

nodedb-sql/src/resolver/derived.rs
EnRaiha: Resolved by the source of truth rather than the comment: WHERE sparse routing is the planner's `Ok(None)` and now declares nothing (`e9f969f60`). A correlated LATERAL subquery cannot be planned at scope time and routes no cells either.

nodedb-sql/src/functions/registry.rs
EnRaiha: `builtin_registry()` is deleted in `e9f969f60`; its only caller was the heuristic. `DEFAULT_REGISTRY` and `FUNCTION_REGISTRY` are untouched.

nodedb/tests/wire/cases/sql_search_subquery_composition.rs
EnRaiha: Pinned in `e0a51d035`: the cell must parse as `f64` with no empty escape; the WHERE operator form that routes a search is asserted numeric, and the comparison form that routes no search asserts `42703`.

nodedb/tests/wire/cases/sql_search_subquery_composition.rs
EnRaiha: Pinned in `e0a51d035`: one outcome per case, and the divergence rows are guards — sparse `WHERE` → `42703`, one-argument `ORDER BY vector_distance(..)` → refusal, a body off a derived relation → refusal, `WHERE multi_vector_search(..)` → declared (lowering refuses `42601`, never `42703`).

PR #339, opened 2026-09-18
fix(query): evaluate derived-table expressions instead of folding to NULL

### LINKED ISSUE #295: Expression errors fold silently over constant derived tables — 22012 raised on collection scans but not on FROM (SELECT ...)

farhan-syah:
Closes #295.

## Symptom

`22012` over `FROM (SELECT 1 AS x) s` folds to NULL. The real class is wider: any computed column, window function, or aggregate over a non-scan derived body (constant, aggregate, grouped, UNION) silently returns NULL, zero rows, or aggregates the base table instead of the derived body.

## Fix

| Area | Change |
|---|---|
| `ProviderScan` / `PostProcess` | Evaluate computed columns and window functions over the derived body |
| `SqlPlan::Subquery` | New `window_functions` field |
| Aggregate | Runs over any derived body via `Aggregate { input }`, materialized on the coordinator |
| Set operations | New `QueryOp::SetOp`; merge logic hoisted out of pgwire into shared code |
| Constants | Typed cells — `SELECT 1` reports `int8` |
| Windows | Partitions order by each window's own `ORDER BY` |
| `OFFSET` | Applied after `ORDER BY` on the provider path |
| Coordinator-local bodies | No longer broadcast to every core |

## Tests

16 new wire tests in `nodedb/tests/wire/cases/sql_subquery_from.rs`. Full suite green at branch head: stage 1 16440, stage 2 369.

## Not in this PR

- `SUM(int)` renders `15.0` on every path
- Native protocol has no merge for top-level `UNION` / `INTERSECT` / `EXCEPT`
- `nodedb-lite` needs a `QueryOp::SetOp` arm and a `SubqueryVisitArgs.window_functions` destructure


farhan-syah:
Folded into #340: one branch, one PR covering #295 and #314.

Issue #338, opened 2026-09-17
Routed response-shaper errors flatten to XX000; the numeric SQLSTATE mapper covers 20 of 81 codes


EnRaiha:
### Version / build tested against

origin/main @ a7f597257

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Document (strict and schemaless), set operations; the pgwire response path (all engines)

### Summary

A response-shaping failure reaches pgwire through one of seven routed call sites that hardcode `sqlstate_error("XX000", ...)`. The shaper's error carries its own numeric classification (`NodeDbError::serialization` → `ErrorCode::SERIALIZATION`, 4200), but the numeric mapper those sites consult covers 20 of the 81 `ErrorCode` constants declared in `nodedb-types` on `main`, and `SERIALIZATION` is among the missing — so even after switching the sites to the mapper, the shaper's own error still renders `XX000`. A client sees an internal fault for a data-representation error.

The same table backs `Error::RemoteTyped` (gateway) and `to_pg_error` (cell encoding), so the gap is not limited to shaping, and the local typed path has the same hole for the same families.

### Related

- #239 — the same loss on the native path: the client's typed error discarded the server SQLSTATE and retriability with it. Closed; this issue is the pgwire-side sibling for responses that never held an `Error` value.
- #306 — a data error reported as `XX000` with a different root cause; the classes added here govern how its verdict renders once it reaches a routed site.
- #266 — misclassification: Calvin infrastructure aborts surfaced to clients as serialization conflicts.
- #313 — class correctness for a missing sequence (`42704`, not `42601`).
- #314 — class choice for sequence accessors (`0A000` by design today).
- #295 — class consistency: `22012` raised on collection scans but folded over constant derived tables.
- #297 — sibling defect on the same response path (kv/timeseries command tags), split into its own change so this mapping fix travels alone.

### Other shape/error surfaces (audited)

The shaper (`response_shape`) is the protocol-neutral layer that turns a Data-Plane payload into rows — KV-wrap, search surrogate translation, payload decode, scan-envelope unwrap, projection, redaction — shared by pgwire, native and HTTP; each protocol only encodes the result. The seven routed pgwire sites are not the only consumers. Callers were enumerated with the code graph (c2g) and cross-checked by source search; the graph drops calls inside closures and match-arm bodies, so both were used together.

| Surface | Site | Current behaviour | Should |
|---|---|---|---|
[...]

PR #341, opened 2026-09-19
fix(pgwire): classify the streamed and DDL shaping failures

EnRaiha:
## Why

Two surfaces still hardcoded the class after the routed sites went through the mapper and the mapper table was completed (PR stacked below):

1. The streamed responder (decode and shape) rendered a class the error already carried as `XX000`.
2. The DDL dispatch (RETURNING shape) did the same.

Part of #338. Top of a two-PR stack, on top of the mapper completion.

## What

- **Streamed surface**: decode and shape failures render through `error_to_sqlstate` with the same `shaping_error_message` hygiene; the two "payload was not a row array" invariants keep `XX000` by decision.
- **DDL dispatch**: the RETURNING shape failure renders through the mapper instead of a literal `XX000`.

## Validation

- `cargo test -p nodedb --lib error_map` — 35 passed.
- `cargo test -p nodedb-types` — 750 passed.
- `cargo clippy -p nodedb -p nodedb-types --all-targets -- -D warnings` — clean.
- Wire tests: not run in isolation for this PR (no wire case changes).

## Notes for review

Read the mapper PR first; this one only moves the two remaining surfaces onto it.


farhan-syah:
Closing: stacked on #359, which conflicts with main, and CI fails here. Resubmit against main once the mapper PR lands.

PR #345, opened 2026-09-19
fix(pgwire): keep the planner's output keys on the extended protocol

### LINKED ISSUE #337: pgwire extended protocol collapses duplicate output column names to one cell

origin/main @ d3d73bed

### Deployment mode

Origin — single node (local)

### Engine(s) involved

Not engine-specific / unsure

### Summary

Over the extended-query protocol (Parse/Bind/Execute), a SELECT with two output columns of the same name renders one cell in both columns. The Describe phase rebuilds the result projection from the announced PG fields with `lookup_key = display_name = field name` (`nodedb/src/control/server/pgwire/handler/prepared/execute.rs`, the `OutputSchema` built from `stmt.result_fields`). That projection overrides the planner's `OutputSchema` at `nodedb/src/control/server/pgwire/handler/routing/execute.rs` (`shaping.projection.or(Some(&output_schema))`). The shaper then reads both cells through the same key. Every producer stores cells under the unique keys from `cell_keys` (`response_shape/project.rs`), so the second column's cell (`<name>_1`) is never read. The simple-query protocol uses the planner schema and is correct after #327.

Found by static review of #327. Derived from the code, not reproduced on a running server.

### Steps to reproduce

```sql
-- Run through a prepared statement (tokio_postgres `client.query`, psycopg `execute`,
-- any driver on Parse/Bind/Execute). `psql` simple-query mode does not reproduce.
CREATE SEQUENCE s START 1 INCREMENT 1;
SELECT nextval('s'), nextval('s');   -- expected (1, 2); returns (1, 1) after #327, (2, 2) before

-- Same class, join with a repeated bare name:
CREATE COLLECTION w (id TEXT PRIMARY KEY, b_id TEXT) WITH (engine='document_strict');
CREATE COLLECTION b (id TEXT PRIMARY KEY) WITH (engine='document_strict');
INSERT INTO b (id) VALUES ('b1');
INSERT INTO w (id, b_id) VALUES ('w1', 'b1');
SELECT w.id, b.id FROM w JOIN b ON w.b_id = b.id;   -- expected (w1, b1); both columns read one cell
```

### Expected behavior

Each output column renders its own cell on every protocol. The lookup keys come from one place: the planner's `build_output_schema`, which already derives per-column unique keys via `cell_keys`. The Describe-built projection must not replace those keys. It can supply display names, types, and result formats only.

### Actual behavior

On Bind/Execute the duplicate columns both read the first stored cell (`nextval` → `1`, `1`). Stored data is intact. Simple-query protocol returns the correct row.

Workarounds: alias each column to a distinct name, or use the simple-query protocol.

### Proposed severity

### Reproducibility

Always — every attempt

Never worked. The override predates `cell_keys`.

### Environment & logs

Linux x86_64. No server log output: the wrong row is returned without error.

### Before submitting

- [ ] I reproduced this on a released tag or a current `main` build (not a stale local branch). — Derived from code reading at `d3d73bed`, not run.

EnRaiha:
## Why

Over Parse/Bind/Execute (extended protocol), two output columns sharing a name render the same cell: `SELECT w.id, b.id …` returns `(w1, w1)` and `SELECT nextval('s'), nextval('s')` returns `(1, 1)`.

The Describe phase built its `OutputSchema` with `lookup_key = display_name = field name` (`prepared/execute.rs:136-149`), and the execute path replaced the planner's columns with it (`routing/execute.rs:153-160`). Rows are written under the unique keys `cell_keys` derives (`id`, `id_1`), so the second column's cell was never read. Simple query has been correct since the derived-key work.

Fixes #337.

## What

- Merge instead of replace: `lookup_key` comes from the planner's `build_output_schema` — the single derivation that runs `cell_keys`; the announced schema supplies `display_name`, `ty`, and `is_star`; `cp_computed` stays the planner's.
- An announced arity that disagrees with the plan keeps the announced columns but derives their keys with `cell_keys`, so the reader still agrees with the row writer.
- `effective_output_schema` is private, unit-tested in place; no signature or public-API change.

## Validation

- Wire (extended protocol) on the harness: `SELECT 1 AS id, 2 AS id` → `(1, 2)`; `SELECT w.id, b.id FROM w JOIN b …` → `(w1, b1)`. Both cases fail on the current `main` (red: `left: 1, right: 2`; the join case reports `error retrieving column 0`) and pass with the fix (green).
- `cargo test -p nodedb --lib routing::execute` — pass (merge keeps the planner keys; arity mismatch derives announced keys).
- rustfmt clean; the repository preflight passes (test attributes +2).

## Notes

- Presentation-only: no authorization, redaction, or audit change; display names, types, and result formats are unchanged.


farhan-syah:
Closing: right behaviour, wrong place.

The producer is `prepared/execute.rs` — the Describe phase builds `lookup_key: f.name()` instead of deriving keys with `cell_keys(names)`. Fix that one site and the existing replace-at-execute stays correct. The 40-line downstream merge is not needed; its arity-mismatch branch already is the producer fix. Resubmit as the producer fix plus the wire test.

PR #347, opened 2026-09-19
feat(lease): fence descriptor-lease release by SWIM verdict and holder incarnation

EnRaiha:
## Why

Two defects in the descriptor-lease release path, both from the P2 consensus-safety review:

1. The drain wait compared a holder's stamped `expires_at` against this node's wall clock. The holder stamps from its own clock and its HLC is never merged in, so a fast local clock could retire a live holder's lease early — the window the drain exists to close.
2. Release waited for topology removal or expiry. A crashed holder stayed a member until the removal path ran, and its leases blocked every DDL drain on those descriptors for the full lease duration.

## What

- **Skew bound.** `lease_expired` gives the stamped deadline the same allowance the HLC ingress enforces (`MAX_CLOCK_SKEW_NS`); a saturated deadline is never retired — the filter drops holds only where the lapse is certain.
- **SWIM-fed liveness.** `LeaseHolderLiveness` (+ `LeaseHolderLivenessHook`): Dead/Left mark a holder, Alive revives (verdicts are refutable), Suspect is a no-op by decision. The GC collects a holder absent from topology **or** marked gone — `collect_non_member_lease_releases(topology, cache, liveness)`. The set is created with the node's cluster handle, registered as a SWIM subscriber at subsystem start (the same topology-backed resolver the routing hook uses), and handed to the `RaftLoop` via `with_lease_liveness`.
- **Incarnation fencing.** `DescriptorLeaseGrantFenced { lease, holder_incarnation }` is appended last (zerompk numbers variants by position) and proposed only under `can_activate_feature(LEASE_FENCING_VERSION)`; mixed-version clusters keep the unfenced grant. `MetadataCache.lease_incarnations` records the stamp; a verdict at `N` releases leases stamped `<= N`, so a restarted node that re-acquired is unaffected; an unstamped lease releases as before.

## Validation

- `cargo test -p nodedb-cluster --lib lease` — 17 passed (collector incl. dead-member, fenced-newer-lease, revive; liveness hooks; clamp boundary).
- `cargo test -p nodedb-cluster --lib metadata_group::cache` — 2 passed (fenced grant records; release clears).
- `cargo check -p nodedb --all-targets` — clean; repository preflight passes.

## Notes

- A SWIM false positive now releases a live node's leases earlier than expiry would. That is the Dead verdict's authority — the same one topology removal already had — and an Alive refutation clears the mark.
- `state/fields.rs` and `state/init.rs` stay at their base line counts: two adjacent doc comments were compacted to make room for the new field.
- Follow-up (not in this PR): a propose-side gate test for the fenced variant.


farhan-syah:
Closing. A SWIM `Dead` verdict does not stop a partitioned-but-alive holder from using its descriptor lease, so releasing on that verdict lets DDL proceed under a live lease. The PR notes the false-positive case, and no holder-side fence covers it. Release must stay expiry-bound (with the skew allowance), unless the holder fences itself first. The maintainers will handle this in-tree.

PR #348, opened 2026-09-19
fix(raft): queue each committed index once per Ready window

EnRaiha:
## Why

The P2 review left one unexplained shape: a Raft batch carrying the same committed index twice. The applier's delivery guard (each committed index handed off at most once, watermark claimed in one critical section) is the boundary — but the producer-side cause was unknown, and an append-shaped write gains a row if that guard ever slips.

Root cause found: `collect_committed_entries` resumed from `last_applied` alone. It runs on every commit-index advance — a follower's AppendEntries, a single voter's propose — while the loop drains `Ready` and advances `last_applied` later. Two advances inside one window queued the same committed range twice, putting one index twice into a single batch.

## What

- `collect_committed_entries` resumes past the furthest index already queued into `Ready` (`max(last_applied, queued_through)`), so a second advance in the window queues only what is new. No new state, O(1).
- `apply_group_commits` records a repeated committed index (`first_duplicate_committed_index`) with a warn — a future producer-side repeat is observable, and the applier guard stays the boundary.

## Validation

- Red proof: the new raft test fails on the base producer (repeated indices in one batch) and passes with the fix.
- `cargo test -p nodedb-raft --lib node::` — 86 passed.
- `cargo test -p nodedb-cluster --lib apply_committed` — 1 passed (detector).
- `cargo check -p nodedb-cluster --all-targets` — clean; repository preflight passes.

## Notes

- A cross-tick re-queue while the async data-plane apply lags remains possible by design (a fresh `Ready` starts empty). That is a repeated delivery across batches, covered by the applier's watermark guard, not a duplicate inside one batch.


nodedb-cluster/src/raft_loop/tick/apply_committed.rs
```diff
@@ -12,10 +12,30 @@ use crate::forward::PlanExecutor;
 
 use super::super::loop_core::{CommitApplier, RaftLoop};
 
+/// The first committed index a batch carries twice, when it carries one.
+///
+/// A batch that repeats an index is the shape that once let an append-shaped
+/// write apply twice before the applier's delivery guard landed. The guard
```
farhan-syah: This narrates a pre-fix state ("once let ... before the applier's delivery guard landed"). Comments state what is. Suggested shape: a batch never repeats an index; the applier guard absorbs one if a producer regresses, and this makes that regression observable.

nodedb-cluster/src/raft_loop/tick/apply_committed.rs
```diff
@@ -12,10 +12,30 @@ use crate::forward::PlanExecutor;
...
+/// stays the boundary; this makes a producer-side repeat observable instead
+/// of silent.
+fn first_duplicate_committed_index(entries: &[nodedb_raft::LogEntry]) -> Option<u64> {
+    let mut seen = std::collections::BTreeSet::new();
+    entries
+        .iter()
+        .map(|entry| entry.index)
+        .find(|index| !seen.insert(*index))
+}
```
farhan-syah: This allocates a `BTreeSet` per applied batch on the tick path. Committed entries are contiguous ascending by construction, so a zero-allocation `windows(2)` strict-increase check is cheaper and the stronger invariant: report the first index that is not greater than its predecessor.

farhan-syah (changes requested):
Root cause verified independently: `collect_committed_entries` is the only producer of `ready.committed_entries`, and the loop calls `advance_applied` only after `apply_group_commits`, so two commit advances between drains re-queue the range. Fix at the producer, no new state. Correct layer.

Claims verified: red proof reproduces on the base producer (`[2, 2, 3]`), `nodedb-raft node::` 86 pass, `nodedb-cluster apply_committed` 1 pass, clippy `--all-targets -D warnings` clean on both crates, fmt clean.

Two inline notes, neither blocks. After the change, amend the commit or squash: one commit on this branch.

nodedb-cluster/src/raft_loop/tick/apply_committed.rs
EnRaiha: Fixed — the doc now states the invariant: committed entries are contiguous and strictly ascending by construction, so a non-increasing pair is a producer regression; the applier's delivery guard stays the boundary, and the detector makes the regression observable. No pre-fix narration remains.

nodedb-cluster/src/raft_loop/tick/apply_committed.rs
EnRaiha: Replaced with `windows(2)`: zero allocation on the tick path, and it reports the first index that is not greater than its predecessor. The test pins `None` for `[]`/`[1]`/`[1,2]`, `Some(2)` for `[2,2,3]` (the red-proof shape), and `Some(1)` for `[1,2,1]`.

farhan-syah (approved):

Issue #349, opened 2026-09-19
bug(ddl): query-function router matches keywords inside string literals and comments


EnRaiha:
## Summary

`query_functions::router::try_dispatch` selects a query function with
`sql.to_uppercase().contains(...)` over the raw statement text. Any occurrence
inside a string literal, quoted identifier, or comment is treated as a call.
Any statement that does not parse as typed DDL and whose text contains one of
the six substrings fails with a misleading `42601`.

- Location: `nodedb/src/control/server/shared/ddl/neutral/query_functions/router.rs:31-48`
- Call path: `shared/ddl/dispatch.rs:33` → `neutral::try_dispatch` → `router::try_dispatch`
- Substrings: `VERIFY_AUDIT_CHAIN`, `VERIFY_HASH_CHAIN`, `BALANCE_AS_OF`,
  `TEMPORAL_LOOKUP`, `VERIFY_BALANCE`, `CONVERT_CURRENCY_LOOKUP`
- The module doc states these checks replicate the pgwire
  `router::function::dispatch` behaviour "verbatim"; that pgwire copy no longer
  exists, so the fix is contained to this router.

## Repro (live server build `3abe3a269`; the file is unchanged on main)

| SQL | Result |
|---|---|
| `INSERT INTO kg_code_nodes (id,name,kind,file) VALUES ('tmp_verify_balance_x','x','function','/x.rs')` | error: `VERIFY_BALANCE requires (collection, column)` |
| `INSERT INTO kg_code_nodes (id,name,kind,file) VALUES ('tmp_plain_x','x','function','/x.rs')` | OK |
| `INSERT INTO kg_code_nodes (id,name,kind,file) VALUES ('tmp_z','verify_balance','function','/x.rs')` (keyword in the label) | error |
| `INSERT INTO kg_code_nodes { id: 'tmp_verify_balance_y', name: 'x', kind: 'function', file: '/x.rs' }` (brace form) | OK |
| `SELECT VERIFY_BALANCE('c','col')` | routes correctly (intended) |

## Impact

- Bulk ingest fails for rows whose id/label/path contains one of the six tokens
  (e.g. `query_functions_verify_balance_file`).
- The error names a function the statement never called.
- A comment such as `-- verify_balance note` can trigger the same hijack.

## Fix plan

1. Add a small scanner that blanks quoted regions (`'...'` with `''` escapes,
   `"..."`) and comments (`--`, `/* */`) before the keyword checks.
[...]

PR #350, opened 2026-09-19
fix(ddl): scan SQL code, not literals, in the query-function router

EnRaiha:
## Why

`query_functions::router::try_dispatch` picked a query function with
`sql.to_uppercase().contains(...)` over the **raw** statement, so any keyword
inside a string literal, quoted identifier, or comment hijacked the statement.
Any statement that does not parse as typed DDL and whose text contains one of
six substrings failed with a misleading `42601`, found while replaying a
knowledge-graph backfill:

| SQL | before |
|---|---|
| `INSERT INTO kg (id) VALUES ('tmp_verify_balance_x')` | `VERIFY_BALANCE requires (collection, column)` |
| `INSERT INTO kg (id) VALUES ('tmp_plain_x')` | OK |
| `INSERT INTO kg (id) VALUES ('tmp_z', 'verify_balance')` (keyword in a label) | hijacked |
| `INSERT INTO kg { id: 'tmp_verify_balance_y', ... }` (brace form) | OK |

Issue: #349.

## What

- add `scan_code`: blanks `'...'` (with `''` escapes), `"..."`, `--`, and
  `/* */` before the keyword checks; all other characters pass through verbatim
- extract `recognized_function` (the routing decision) so it is testable;
  `try_dispatch` matches on it; recognition order unchanged; a real
  `SELECT VERIFY_BALANCE(...)` still routes

## Tests

- unit tests in `query_functions::router` now pin the **routing decision**
  (`recognized_function`): literal, escaped quote, quoted identifier, line and
  block comment, token fusion, first-keyword order, and real calls
- wire regression in `nodedb/tests/wire/cases/router_misroute_literals.rs`
  (`value_carrying_verify_balance_does_not_misroute`): a parenthesised INSERT
  whose value carries the token stores verbatim — the brace form is intercepted
  before the router, so it cannot guard this — and the anchored call still
  reaches the function arm
- local: `cargo test -p nodedb --lib query_functions::router` -> 9 passed;
  `cargo test -p nodedb --test wire value_carrying_verify_balance_does_not_misroute`
  -> 1 passed
- CI: static gates pass; the `Test Suite` workflow runs only with the `run-ci`
  label (`.github/workflows/ci.yml` gates it), so it reports `skipping` by design

Survey with c2g on the branch: diff-impact = the decision helper and its call
site; no exported surface change.

## Notes

- Scanner scope: single-quoted literals (`''` escapes), quoted identifiers
  (`""`), `--` and `/* */` comments. Not handled: nested block comments,
  E-string backslash escapes, and dollar-quoted strings — pre-existing class,
  not regressions of this change.
- Follow-up, separate refactor: dedupe with the private scanners in
  `planner/procedural/executor/core/sql_literal_concat.rs`
  (`consume_quoted_region`, `parse_single_quoted_literal`) — requires a
  visibility change.


farhan-syah:
Closing. The fix keeps routing by substring and adds a second quote/comment scanner beside the existing one. The correct fix routes on the parsed function name, so literals, identifiers, and comments cannot match at all. The maintainers will fix #349 in-tree.

PR #351, opened 2026-09-19
feat(vector): fused BBQ dequantize-distance kernels with tiered SIMD dispatch

### LINKED ISSUE #280: Fused BBQ dequantize-distance kernels with tiered SIMD dispatch
### Problem
Per-candidate rerank distance for 1-bit BBQ codecs allocates a `Vec<f32>` reconstruction per candidate (encode → dequantize → subtract → L2). For a 768-dim collection at `oversample × ef` candidates this is the dominant rerank cost and defeats SIMD.

### Proposed solution
A fused kernel that computes `dist² = Σq² − 2·scale·Σ(sign·q) + dim·scale²` directly over packed sign bits, dispatched through a small function-pointer table chosen at runtime:
- AVX-512 (512-bit server: Ice Lake+, Zen 4/5) — 32 lanes, `_mm512_mask_blend_ps` / masked ops
- AVX10/AVX-512VL (256-bit hybrid: Sierra/Clearwater Forest, Core Ultra) — `_mm256_mask_blend_ps`, `__mmask8`
- AVX2+FMA (Haswell..Alder Lake, Zen 1-3) — `_mm256_blendv_ps`
- NEON (aarch64) — `vbslq_f32`, with a `vtbl1_u8` LUT for bit expansion
- WASM SIMD128 — `v128_bitselect`
- Portable scalar fallback (zero-allocation)
Bit order is MSB-first; `byte.reverse_bits()` maps dim k → lane k. Correctness gate: golden tests vs scalar across dims 8..512, tolerance 1e-4; criterion benches vs the current alloc path.

### Engine / area
Vector (rerank codecs).

### Alternatives & workarounds
Current code path is correct, just slower; scale by reducing oversample.

### Before submitting
- [x] I searched existing issues and discussions for this proposal.

EnRaiha:
## Why

Rerank against 1-bit BBQ candidates materialized a `Vec<f32>` per candidate
(and per query). At an `oversample × ef` candidate count that dominated the
pass and defeated SIMD. This is the fused-kernel work of #280.

## What

- the kernels read the centered query **straight from the prepared payload
  bytes**: zero allocation per candidate and per query
- runtime dispatch through a function-pointer table resolved once:
  `avx512` → `avx10-256` → `avx2` → `neon` → `wasm-simd128` → `scalar`
- every tier derives its lane mask with the same MSB-first mapping
  (`byte.reverse_bits()`, dim `k` → lane `k`), and every tail is handled by the
  scalar accumulator so a vector tail cannot diverge from the head formula
- `decode_payload` / the old dequantize path are removed

## Validation

- golden parity against an f64 oracle over dims 0..768 — the issue's 8..512
  range plus lane boundaries and partial tails; relative tolerance 1e-4 with an
  absolute floor of 1e-6
- native (`avx2` + `fma`): `cargo test -p nodedb-vector --lib rerank::codecs`
  → **38 passed**
- the 512-bit tiers have no native hardware in this fleet; they run under
  **Intel SDE 10.8.0** (`sde64 -spr`, the release rust-lang pins in stdarch CI):
  `rerank::codecs::bbq_kernel` → **6 passed**, including
  `avx512_tier_matches_the_reference` and `avx10_256_tier_matches_the_reference`
- aarch64: `cargo check --target aarch64-unknown-linux-gnu` clean (NEON tier
  compiles)
- cross-target: aarch64 NEON tier **executed under `qemu-aarch64`**
  (`neon_tier_matches_the_reference` passes; runner:
  `CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc` +
  `…_RUNNER="qemu-aarch64 -L /usr/aarch64-linux-gnu"`), and `cargo check`
  clean; the wasm tier compiles
  clean for `wasm32-wasip1` with `-C target-feature=+simd128` (the
  `wasm32-unknown-unknown` check is blocked by a pre-existing `getrandom` graph
  issue). The wasm check surfaced two real tier bugs, both fixed here: a missing
  `unsafe` block on `v128_load`, and an unreachable scalar arm on wasm targets.
  A standalone probe of the same kernel shape (dim 768, Node 22/WASI) measures
  the wasm tier at 1,384 ns/op vs 6,817 ns/op for wasm-scalar (4.9x) and 4.0x
  behind native AVX2 — the expected cost of 4-lane vs 8-lane arithmetic and no
  FMA in SIMD128 (the wasm tier is a portability tier, not a speed tier)
- zero allocation is proven, not asserted: with fluxbench's `TrackingAllocator`
  installed, both fused benches report **0 bytes / 0 allocations**, while the
  replaced path allocates one `Vec<f32>` per candidate (131,072 B per
  256-candidate iteration at dim 128; 786,432 B at dim 768)
- benches against the replaced path (fluxbench, the repo norm): fused is
  **7.3×** (dim 128) and **10.3×** (dim 768) faster with allocation tracking
  installed (13.6× / 16.0× without it)
- `cargo clippy -p nodedb-vector --lib` → 0 warnings; preflight pass

## Notes / deviations from the issue text

- the issue says "32 lanes" for AVX-512; for f32 `__m512` holds **16 lanes**
  (`__mmask16`), which is what the kernel uses
- benches use the repo's `fluxbench` harness, not criterion
- AVX2 uses a static sign-lane table + FMA rather than `_mm256_blendv_ps`, and
  NEON uses `vtstq_u32` weight vectors rather than `vtbl1_u8` — same MSB-first
  mapping, documented in the code
- the `avx10-256` arm is reserved for AVX10/256 hosts: AVX512VL implies
  AVX512F, so today it is exercised by its direct test rather than by
  `selected_kernel()`

## Known gaps (not covered by this PR)

- AVX-512 / AVX10 correctness is validated under Intel SDE only; there is no
  native 512-bit hardware here, so no real-silicon performance numbers
- the `avx10-256` dispatch arm cannot be selected at runtime today (AVX512VL
  implies AVX512F); its correctness is pinned by its direct test until AVX10
  detection lands in `std::arch`
- the crate's own wasm unit test cannot build yet: `tokio` (via the fluxbench
  dev-dependency) does not build for wasip1. The wasm tier is compile-checked
  and executed through the standalone probe; a wasm test job needs
  target-gated dev-deps or `wasm-bindgen-test`
- no per-tier benches: only the dispatched tier is measured natively (the wasm
  figure comes from the probe)
- differential testing uses fixed seeds; there is no fuzz/property target for
  the kernel yet

## Testing the 512-bit tiers locally (no 512-bit hardware needed)

Intel SDE emulates the 512-bit paths — it is the same release rust-lang pins in
its stdarch CI:

- SDE page: https://www.intel.com/content/www/us/en/developer/articles/tool/software-development-emulator.html
- Direct download: https://ci-mirrors.rust-lang.org/sde-external-10.8.0-2026-03-15-lin.tar.xz

```bash
curl -LO https://ci-mirrors.rust-lang.org/sde-external-10.8.0-2026-03-15-lin.tar.xz
tar xf sde-external-10.8.0-2026-03-15-lin.tar.xz -C /opt
CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="/opt/sde-external-10.8.0-2026-03-15-lin/sde64 -spr --" \
  cargo test -p nodedb-vector --lib rerank::codecs::bbq_kernel
# expect 6 passed, including avx512_tier_matches_the_reference and
# avx10_256_tier_matches_the_reference; chip flags are short: -spr -emr -gnr -cwf
```

Closes #280


nodedb-vector/src/rerank/codecs/bbq_kernel.rs
```diff
@@ -0,0 +1,612 @@
...
+        "centered payload {} < dim*4 {}",
+        centered.len(),
+        dim * 4
+    );
+    selected_kernel().distance(centered, packed, residual_norm, dim)
+}
+
+/// The tier chosen for this host; resolved once.
+pub(super) fn selected_kernel() -> &'static Kernel {
+    static KERNEL: OnceLock<Kernel> = OnceLock::new();
```
farhan-syah: This is the second dispatch system: its own `OnceLock` and its own feature detection, beside `SimdRuntime` in `distance/simd/runtime.rs`. Register the BBQ kernel as a `SimdRuntime` field, selected in `detect()`, the same way `l2_squared_f16`/`l2_squared_bf16` are.

nodedb-vector/src/rerank/codecs/bbq_kernel.rs
```diff
@@ -0,0 +1,612 @@
...
+                name: "avx512",
+                // SAFETY: guarded by the feature detection above.
+                l2: |c, p, n, d| unsafe { l2_avx512(c, p, n, d) },
+            };
+        }
+        // Reserved for AVX10/256 hosts: AVX512VL implies AVX512F, so on
+        // today's feature bits this arm is reachable only when AVX10 detection
+        // lands in std::arch (the 256-bit masked kernel matches the AVX10.1
+        // shape). Its correctness is pinned by `avx10_256_tier_matches_the_reference`.
+        if std::is_x86_feature_detected!("avx512vl") && std::is_x86_feature_detected!("avx512f") {
```
farhan-syah: This arm can never be selected. `avx512vl` implies `avx512f`, so any host that passes this check already returned the `avx512` tier above. The comment says so too. A tier that production cannot reach is dead code in the dispatch. Remove it, or gate it on a real AVX10 detection once `std::arch` has one.

farhan-syah (changes requested):
The kernel work is strong and carries real evidence: a parity check against an f64 reference over dims 0..768, the 512-bit tiers run under Intel SDE, NEON run under qemu, and allocation tracking that shows zero bytes. That part stays.

**Blocker: a second SIMD dispatch system.** `nodedb-vector` already selects kernels once, through `distance/simd/runtime.rs` (`SimdRuntime`, `runtime()`). It also already holds fused kernels that decode and compute without an intermediate `Vec<f32>`: `l2_squared_f16` and `l2_squared_bf16`. That is the exact precedent for this kernel. This PR adds its own `OnceLock`, its own feature detection, and its own kernel table under `rerank/codecs`. Two dispatch systems can pick different tiers for one host, and each new kernel must then choose which one to join.

Direction:
- Add the BBQ distance as a `SimdRuntime` field, selected in `SimdRuntime::detect()` next to the f16/bf16 kernels.
- Move the per-tier kernels beside their siblings in `distance/simd/{avx512,avx2,neon}.rs`.
- The rerank codec calls `runtime().<field>`.

**Should-fix: an unreachable tier** (inline).

After the change, rebuild the commits rather than appending fix-up commits on top.

EnRaiha:
Reworked and rebuilt into one commit (`f4b973cfd`), based on `main @ 1ff35512b`.

| Point | Landed |
|---|---|
| **Blocker: a second dispatch system** | The `OnceLock`, `detect_kernel`, and the `Kernel` table are gone; `feat/280-bbq-fused` now adds `pub type BbqFn` and `pub l2_bbq: BbqFn` to `SimdRuntime`, with one selection line per tier arm in `detect()`. `grep -rn "OnceLock\|detect_kernel\|bbq_kernel" nodedb-vector/src` returns only the pre-existing `static RUNTIME` |
| Per-tier kernels beside their siblings | `distance/simd/{avx512,avx2,neon}.rs` carry `l2_bbq` (safe entry) + `l2_bbq_impl` (`#[target_feature]`); the wasm kernel is in `distance/simd/wasm_simd128.rs`; the scalar reference and the shared helpers are in `distance/simd/bbq.rs` |
| The rerank codec calls `runtime().<field>` | `rerank/codecs/bbq.rs:178` — `(crate::distance::simd::runtime().l2_bbq)(…)` |
| **Should-fix: an unreachable tier** | Removed, with its test. No `avx10_256` remains; the comment promising AVX10 detection went with the arm |
| Rebuild the commits | One commit, message rewritten to describe the final shape |

Evidence, on the rebased tree:

| Check | Result |
|---|---|
| `cargo test -p nodedb-vector --profile ci bbq` | 18 passed, 0 failed (was 19 — the difference is the removed tier's test) |
| `cargo test -p nodedb-vector --profile ci --lib` | 433 passed, 0 failed |
| `cargo check --all-targets` | 0 warnings, 0 errors |
| `-C target-feature=+simd128` for `wasm32-wasip1` | clean — the wasm kernel compiles |
| Parity | unchanged: the f64 oracle over dims 0..768, plus `dispatch_selects_a_tier_that_matches_the_reference` over the whole table |

Two things I fixed in passing that the review would have caught: the whole-range scalar kernel needs `.sqrt()` (the tier kernels return a distance, the accumulation returns a squared sum), and `#[target_feature]` belongs on the unsafe impl rather than the safe wrapper — both now match the f16/bf16 shape.

Not run here: the AVX-512 tier under Intel SDE and NEON under qemu (this host is AVX2-only; the avx512 test prints its skip line), and the bench numbers in `benches/bbq_kernel.rs` were not re-measured.

nodedb-vector/src/distance/simd/bbq.rs
```diff
@@ -0,0 +1,228 @@
...
+            centered.extend_from_slice(&x.to_le_bytes());
+        }
+        let packed: Vec<u8> = (0..dim.max(1).div_ceil(8))
+            .map(|_| (next() % 256) as u8)
+            .collect();
+        let residual_norm = ((next() % 1000) as f32 / 100.0) + 1.0;
+        (centered, packed, residual_norm)
+    }
+
+    /// The issue's acceptance range: every dim in `DIMS` against the oracle.
```
farhan-syah: Leak again

nodedb-vector/src/distance/simd/bbq.rs
```diff
@@ -0,0 +1,228 @@
...
+        fn check_tier(name: &str, f: fn(&[u8], &[u8], f32, usize) -> f32) {
+            for dim in DIMS {
+                let (centered, packed, residual_norm) = sample(dim, 1234 + dim as u64);
+                let expected = reference_l2(&centered, &packed, residual_norm, dim);
+                let got = f(&centered, &packed, residual_norm, dim);
+                assert!(
+                    within_parity(got, expected),
+                    "{name} dim {dim}: {got} != reference {expected}"
+                );
+                // The issue asks for parity against the scalar kernel directly.
```
farhan-syah: Leak

nodedb-vector/src/rerank/codecs/bbq_kernel.rs
EnRaiha: The second dispatch system is gone. Its `OnceLock` and kernel table are removed, and `rerank/codecs/bbq_kernel.rs` no longer exists.

`l2_bbq` is now a `SimdRuntime` field, `pub l2_bbq: BbqFn`, selected once per tier arm in `detect()` beside `l2_squared_f16` / `l2_squared_bf16`. The per-tier kernels moved to `distance/simd/{avx512,avx2,neon,wasm_simd128}.rs`, the scalar reference to `distance/simd/bbq.rs`, and the codec calls `(crate::distance::simd::runtime().l2_bbq)(…)` at `rerank/codecs/bbq.rs:178`.

`grep -rn "OnceLock\|detect_kernel\|bbq_kernel" nodedb-vector/src` now matches only the pre-existing `static RUNTIME`. Rebuilt into one commit, `f4b973cfd`. Full evidence table in the timeline comment above.

nodedb-vector/src/rerank/codecs/bbq_kernel.rs
EnRaiha: Removed, along with its test. `git grep avx10` over `nodedb-vector/` returns nothing, and the comment promising a later AVX10 gate went with the arm. The tier returns when `std::arch` can detect AVX10.

EnRaiha:
One more fix before re-review, pushed as an amended commit (`7f3998407`, rebased shape unchanged): an audit of the wasm tier found it had the same defect you flagged on the `avx10-256` arm — a tier that compiles and is tested but that production could not select.

`detect()` had no `wasm32` arm, so a wasm32 + simd128 build fell through to the scalar fallback (`l2_bbq: bbq::l2_bbq`). The `simd128` kernel was reachable only from its own test, which calls the tier module directly rather than the dispatch table.

Fixed in `distance/simd/runtime.rs`:

- the arm now installs `wasm-simd128` when targeting wasm32 with simd128, placed after `neon` and before the scalar fallback
- its guard is the same `target_feature = "simd128"` condition as the tier module, since wasm has no runtime feature probe to consult

This is 25 lines in one file. The arm is verified by execution, not by compilation alone: a probe built for `wasm32-wasip1` under Node 22's WASI runner reports `dispatch name: wasm-simd128` and matches the f64 oracle across 14 dims (1, 7, 8, 9, 63, 64, 65, 127, 128, 129, 255, 256, 512, 768) when calling through `rt.l2_bbq` — the dispatched pointer, not the tier directly.

On the rebased tree:

| Check | Result |
|---|---|
| `cargo test -p nodedb-vector --profile ci bbq` | 18 passed, 0 failed |
| `cargo test -p nodedb-vector --profile ci --lib` | 433 passed, 0 failed |
| `cargo clippy -p nodedb-vector --lib` | 0 warnings |
| `cargo check -p nodedb-vector --all-targets` | 0 warnings, 0 errors |
| `--target aarch64-unknown-linux-gnu` | clean |
| `--target wasm32-wasip1` + `+simd128` | clean |
| `--target wasm32-wasip1` without `+simd128` | clean — the arm and its import gate out, scalar fallback stands |

Two things to flag plainly:

1. This is a force-push, so the inline threads you opened now read as outdated — the file they were anchored to, `rerank/codecs/bbq_kernel.rs`, no longer exists and the commit they pointed at is gone. Both are answered in this thread.
2. The bench figures in the description (7.3× / 10.3×) are carried over from the pre-rework build and are marked as not re-measured. The rework changed dispatch wiring and module layout, not the kernel bodies or the bench harness.

The description has been rewritten to match the current tree. Ready for another look.

PR #359, opened 2026-09-20
fix(pgwire): complete the shaper error classification and keep internal detail in the log

EnRaiha:
## Why

`numeric_code_to_sqlstate` answered 22 of 81 codes and fell back to `XX000` for the rest, so a quota, clone, move-tenant, mirror, sync, storage/WAL, cluster, config, or encryption fault reached every routed surface as an internal fault — including the shaper's own `SERIALIZATION` code, the exact "payload does not fit the projection" case the class should name.

An `XX000`-class failure also rendered its message verbatim, so a manifest path or invariant text could travel to a client on an error frame.

Part of #338. Bottom of a two-PR stack.

## What

- **Mapper completion** (`numeric_code_to_sqlstate`, 22 → 81): the remaining 59 codes carry issue #338's decided class — the constraint family mirrors the Data Plane table (`236xx`); clone, quota, move-tenant, mirror, sync, storage/WAL, cluster, config, and encryption follow the classes the local `crate::Error` table already names. `INTERNAL`, `BRIDGE`, and `DISPATCH` keep `XX000` as explicit arms.
- **Six new `sqlstate.rs` constants**: `22000`, `22P02`, `25006`, `2BP01`, `42P07`, `XX001`.
- **Hygiene**: new `shaping_error_message`; an `XX000`-class failure logs the detail with its code and sends one stable summary, every other class passes through.

## Validation

Run at the head of the stack (this PR plus the surface PR above it):

- `cargo test -p nodedb --lib error_map` — 35 passed (mapper table covering all 59 completion codes, the three internal arms, hygiene replace/pass-through).
- `cargo test -p nodedb-types` — 750 passed (sqlstate constants incl. the six new ones).
- `cargo clippy -p nodedb -p nodedb-types --all-targets -- -D warnings` — clean.

## Notes

The streamed responder and the DDL dispatch render through this mapper in the PR stacked above.


EnRaiha:
Rebased onto `main @ 4664c8213` and pushed (`2f8f4f17c`). The single conflict was in `error_map.rs`, where main's new `Error::Shaping` arm landed: that arm is kept and the fallback bridge follows it. Local evidence: `cargo test -p nodedb --lib error_map` — 37 pass. The branch is linear again (3 commits over main), so the streamed/DDL surfaces work can be resubmitted against main once this lands.

farhan-syah:
Closing. Several classes in the completed table are wrong: `SERIALIZATION` → `22P02` reports internal encode failures as client input errors, `STORAGE`/`WAL` → `58030` renders raw detail (paths) to the client, and `CLUSTER` → `08006` tells drivers the connection dropped. The branch also conflicts with main after #360. The maintainers will settle the SQLSTATE table for #338 in-tree.

